PR #6, opened 2021-10-29
✨ CJS sample package for testing

dmihalcik-virtru:
I haven't wired this in to the makefile since nvm doesn't play nice with make


dmihalcik-virtru:
We should move this to documentation or some other folder; it doesn't make sense here in among the recommended and supported ways to use this

dmihalcik-virtru:
We should move this to documentation or some other folder; it doesn't make sense here in among the recommended and supported ways to use this

PR #33, opened 2022-04-14
webpack builded, working both on Node and Browser

ivanovSPvirtru:


lib/prepareNodeESM.js
```diff
@@ -0,0 +1,33 @@
...
+      }
+      return files
+    });
+
+  files.forEach((filepath) => {
+    fs.readFile(filepath, 'utf8', (err, data) => {
+      if (!data.match(/(?:import|export) .* from/g)) {
+        return
+      }
+      let newData = data.replace(/((?:import|export) .* from\s+['"])([./].*)(?=['"])/g, '$1$2.js')
```
dmihalcik-virtru, line 23: Can you allow multi-line import/export statements by matching on `\n((?:import|export)[\s][^;]*[\s]from\s+['"])([.][/-$\d\w]+)(?=['"])/g, '$1$2.js')`

I think this will work? I haven't tested it, so probably not...

Also, when I did this in the past I did some work to not accidentally double-add `.js` to the end, but I don't remember how to do that...

This will avoid conflicts with prettier and other auto-formatters

lib/src/index.node.ts
```diff
@@ -0,0 +1,13 @@
+import { NanoTDFClient, NanoTDFDatasetClient, AuthProviders, version, clientType } from './index';
+import fetch from 'node-fetch';
```
dmihalcik-virtru, line 2: Yeah, it looks like node-fetch is the way to go:

https://www.npmtrends.com/cross-fetch-vs-isomorphic-fetch-vs-node-fetch-vs-whatwg-fetch

lib/src/index.node.ts
```diff
@@ -0,0 +1,13 @@
...
+// @ts-ignore
+import crypto from 'crypto';
+
+// @ts-ignore
+globalThis.crypto = crypto;
```
dmihalcik-virtru, line 7: Is this safe? I was worried about polluting globals so I avoided it in my node PR.

I didn't bother with fetch though :-/

dmihalcik-virtru, line 7: Is this safe? I was worried about polluting globals so I avoided it in my node PR.

I didn't bother with fetch though :-/

dmihalcik-virtru (changes requested):
I worry without multiline regex this is going to inevitably cause weird build bugs

lib/src/index.node.ts
ivanovSPvirtru: I think it is cause its explicit node.js endpoint and node crypto stuff is goes trough import, not global object. if you want we can use for preventing even possibility of colition?

lib/prepareNodeESM.js
ivanovSPvirtru: Because of it runs trough transpiled stuff its ok, tsc removes brakes
image

dmihalcik-virtru: Okay

lib/src/index.node.ts
dmihalcik-virtru: I'll follow up if we need to. We can discuss it with the channel

dmihalcik-virtru (approved):

lib/prepareNodeESM.js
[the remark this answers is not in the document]
ivanovSPvirtru: Updated too not duplicate the js. Line breaks - i cant fix for some reason. But good news is type script compilers remove line breaks no matter how long they are. So line breaks are not the issue

PR #39, opened 2022-04-21
🔥 Remove some crypto compatibility layer

dmihalcik-virtru:
- Removes support for Safari <11, IE provided with `getCrypto` filler. Global usage <= 0.04%
Safari 7-10.x Global usage <= 0.04%

- Improves node poly filler to avoid re-setting existing poly fills of crypto or fetch


dmihalcik-virtru:
@cassandrabailey293 Is it okay to drop support for these older browsers (IE and safari <= march 2017?)

lib/src/index.node.ts
```diff
@@ -2,8 +2,8 @@ import { NanoTDFClient, NanoTDFDatasetClient, AuthProviders, version, clientType
 import fetch from 'node-fetch';
 import { webcrypto as crypto } from 'node:crypto';
 
-globalThis.crypto = crypto as unknown as Crypto;
-globalThis.fetch = fetch as typeof globalThis.fetch;
+globalThis.crypto ??= crypto as unknown as Crypto;
```
pflynn-virtru, line 5: no idea what `??=` does

pflynn-virtru (approved):

lib/src/index.node.ts
dmihalcik-virtru: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment

to avoid overwriting other library's poly fills

PR #40, opened 2022-04-22
Added exported modules to globalThis and included more build scripts

danielRicaud:
…PLAT-1638

# Conflicts:
#	cli/src/cli.ts


dmihalcik-virtru:
If we want to do this for the web pack bundled build, which kinda makes sense, do this as a new entry point, akin to index.node.ts, so we don't clutter the primary web entry point with globals and side effects

PR #41, opened 2022-04-25
🐛 Fix incorrect dataset i.v. guard

dmihalcik-virtru:
- We are looking for 3 bytes, not 3 nibbles


dmihalcik-virtru:
This is blocked by https://github.com/opentdf/client-web/pull/37, if anyone wants to review that first 🙏

PR #42, opened 2022-04-26
✅ round trip test, less rewrap

dmihalcik-virtru:
- Adds a round-trip test for nanotdf encrypt/decrypt
- like the 'roundtrip' tests in tdf3-js, this mocks out the entirety of the key rewrap mechanism, which is arguably the harder bit
- removes the unused unwrappedKey field. Really, we need to do an audit to make this more functional overall and less state-preserving
- Adds back source maps for better error reports. 
- We should build on top of this somehow to add more coverage of unusual situations

### DIFF
--- a/lib/src/index.ts
+++ b/lib/src/index.ts
@@ -16,29 +16,28 @@ import { TypedArray, createAttribute, Policy } from './tdf/index';
  *
  * @example
  * ```
- * import NanoTDFClient from '@opentdf/client';
+ *  import { clientSecretAuthProvider, NanoTDFClient } from '@opentdf/client';
  *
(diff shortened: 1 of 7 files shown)


dmihalcik-virtru:
Comparison of coverage numbers:
Before:
image

After: 
image

dmihalcik-virtru:
I'm guessing line coverage went down since I turned on source maps, which probably could include a bunch of typescript declarations?

ysizarie-virtru (approved):

danielRicaud (dismissed):

danielRicaud (approved):

PR #47, opened 2022-04-28
Tdf3 + Nano = ❤️

ivanovSPvirtru:


dmihalcik-virtru:
Is this ready to review? It doesn't look like it builds yet, and it isn't clear how to initialize/use/test it. Maybe add a test scenario to the CLI tool?

lib/tdf3/karma.conf.cjs
```diff
@@ -0,0 +1,133 @@
+// Karma configuration
```
dmihalcik-virtru, line 1: Is this file still used?

lib/tdf3/README.md
```diff
@@ -0,0 +1,344 @@
+[![CI Status](https://github.com/opentdf/tdf3-js/actions/workflows/build.yaml/badge.svg)](https://github.com/opentdf/tdf3-js/actions/workflows/build.yaml)
```
dmihalcik-virtru, line 1: Is this file used by typedoc or something? You can probably just delete it otherwise

lib/fasade/FileClient.ts
```diff
@@ -0,0 +1,67 @@
+import { Client as ClientTdf3 } from '../tdf3/src/client';
```
dmihalcik-virtru, line 1: It is spelled `facade`. But this is great! 👍

dmihalcik-virtru (approved):
It looks good, but you'll probably want a clean run of `cd lib && npm test` in the GitHub actions. I'm guessing there is still some cleanup there.

I'm guessing we'll need some more extreme refactoring to get the AuthProvider class working with the old code, so I'm fine with committing it without that now. 

Probably we should commit this as soon as we can get it building and unit tests passing, then we can shard out the various refactoring tasks easier, such as any node->web conversions, addition of types, or changes to auth.

dmihalcik-virtru (approved):

PR #48, opened 2022-05-02
WIP generic oidc client?

dmihalcik-virtru:
- tilt up, then update abacus-web client in keycloak to allow redirects to localhost:8080


dmihalcik-virtru:
I doubt anyone needs this; we can move it to a blog post or something

PR #54, opened 2022-05-19
📄 License check update

dmihalcik-virtru:
- CC-BY-4 is an OSI approved OSS license that meets the other criteria we have for MIT and BSD licenses.
- The only package we transitively depend on that uses it is already allowed

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@opentdf/cli",
-  "version": "0.1.1",
+  "version": "0.2.1",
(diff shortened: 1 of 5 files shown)


cassandrabailey293:
LGTM - there's a conflict in the `lib/package.json` due to https://github.com/opentdf/client-web/pull/55 that needs resolving but otherwise looks fine

dmihalcik-virtru:
Got an approval from PM; can I get a dev review? (@ivanovSPvirtru or @danielRicaud maybe?)

obidenko (approved):

PR #55, opened 2022-05-19
🏷 typescript enhancements

dmihalcik-virtru:
- More return types
- fewer ts-ignores
- some more types!

### DIFF
--- a/lib/tsconfig.json
+++ b/lib/tsconfig.json
@@ -25,5 +25,5 @@
     "strictPropertyInitialization": true,
     "target": "ES2017"
   },
-  "include": ["src", "test"]
+  "include": ["facade", "src", "test", "tdf3"]
 }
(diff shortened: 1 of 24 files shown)


lib/tsconfig.json
```diff
@@ -25,5 +25,5 @@
     "strictPropertyInitialization": true,
     "target": "ES2017"
   },
-  "include": ["src", "test"]
+  "include": ["facade", "src", "test", "tdf3"]
```
pflynn-virtru, line 28: it would be nice to remove test from bundle

pflynn-virtru (approved):

lib/tsconfig.json
dmihalcik-virtru: This is for running tsc. Bundling is accomplished with web pack.config, which I think doesn't include these? (It gives an entry point, and then tests/etc should be shaken out), or the package.json `files` directive, which removes them from the esm module that is the only thing that seems to be built.

PR #58, opened 2022-06-10
Feature/plat 1613 webstream

ivanovSPvirtru:


lib/tdf3/src/client/BrowserTdfSteam.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  static async toBuffer(stream) {
+    const reader = stream.getReader();
+    let accumulator = new Uint8Array();
+    let done = false;
+
+    while (!done) {
+      const result = await reader.read();
+      if (result.value) {
```
dmihalcik-virtru, line 34: Sorry I'm rabbit holing on this. This is a O(n²) algorithm, so I'd rather not do this. An ideal algorithm is probably something like https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/util/ArrayList.java#l237.

However, it roughly the same worst case space performance as just creating a list so I'm inclined to do that instead. This would then be two pass algorithm - first build a standard array of TypedArrays and count the bytes, then allocate memory and assign in a second loop

Alternatively if you want to save a bit of memory you can try some hybrid or space aware approach.

Also does it really need to be a Buffer buffer instead of something that implements the Buffer interface? An alternative is to just implement a 'fractured' buffer but I'm guessing that isn't something we can entertain at the moment

PR #63, opened 2022-07-11
🔥 Remove forked auth classes

dmihalcik-virtru:


ivanovSPvirtru (approved):
Nothing in this world i love more than removing excessive cod

PR #68, opened 2022-07-20
🔥 Remove 'keycloak realm' as param

dmihalcik-virtru:
- Previously, we explicitly required passing in 'realm', variously referred to as `orgName` or `organizationName`, which we then used to build a keycloak-style URL
- This is not part of the OIDC spec and a keycloak addition to support multi-tenancy. This is evident in how all realm requests are based off the `[root]/realms[realm]` path, including the `.well-known` OIDC discovery paths.
- Instead, explicitly include the path as part of the root OIDC url
- While I'm here, remove a few explicit references to keycloak


bleggett:
Yeah I did this but was never terribly happy with it.

All OIDC providers/IdPs (Okta, etc) that are multitenant (that is, all of them) _will_ have a similar construct (org, realm, etc) - but yeah I see no real reason why we can't just make it part of the URL.

bleggett (dismissed):
LGTM

pflynn-virtru (dismissed):

pflynn-virtru (approved):

PR #76, opened 2022-08-18
🐛 Fix a ReferenceError in node

dmihalcik-virtru:
When loading the generated cjs in node contexts, the relevant code generates a value error on dereference. Checking for undefined explicitly should prevent this behavior.

### DIFF
--- a/lib/tdf3/test/setup.js
+++ b/lib/tdf3/test/setup.js
@@ -2,8 +2,10 @@
 import { webcrypto } from 'node:crypto';
 import { ReadableStream } from 'node:stream/web';
 
-if (!globalThis.ReadableStream) {
+if (typeof globalThis.ReadableStream === 'undefined') {
   globalThis.ReadableStream = ReadableStream;
 }
 
-globalThis.crypto = webcrypto;
+if (typeof globalThis.crypto === 'undefined') {
+  globalThis.crypto = webcrypto;
+}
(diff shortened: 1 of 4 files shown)


lib/tdf3/test/setup.js
```diff
@@ -2,8 +2,10 @@
 import { webcrypto } from 'node:crypto';
 import { ReadableStream } from 'node:stream/web';
 
-if (!globalThis.ReadableStream) {
+if (typeof globalThis.ReadableStream === 'undefined') {
```
cassandrabailey293, line 5: any logical inequivalency that will bite us here? for example, a time where we'd like this to evaluate to `true` besides when `typeof globalThis.ReadableStream === 'undefined'` ?

dmihalcik-virtru: This is used in the jest/node setup code so I'm not too worried about it.

The other call in tdf3/src/crypto/index.ts is misleading, as some runtimes will have crypto defined but missing some algorithms or other features. If we really want this feature we should do some checks to make sure that the requested algorithms are supported. For example, we might find a browser that supports TDF3 but not nano. For now, I'm content to avoid the syntax error so the code at least loads in nodeJS and we can add true feature detection if required later

cassandrabailey293 (approved):
LGTM!

danielRicaud (approved):

PR #77, opened 2022-08-22
Handle case when safari return object with stream

krokhmaliuk-virtru:


lib/facade/FileClient.ts
```diff
@@ -79,6 +79,15 @@ export class FileClient {
     if (params) {
       return await this.client.decrypt(params);
     }
+
+    //@ts-ignore
+    const stream = source?.stream;
```
ivanovSPvirtru, line 84: image
If its not PlainTextstream it will be undefined. And decrypt can use other sources.
And remove ts-ignore, it was probably helping you by highlighting that

PR #79, opened 2022-08-25
Fixed crypto import and isNodeStream function

ivanovSPvirtru:


lib/facade/FileClient.ts
```diff
@@ -14,7 +14,8 @@ interface FileClientConfig {
 }
 
 function isNodeStream(source: InputSource): source is NodeJS.ReadableStream {
-  return Object.prototype.hasOwnProperty.call(source, 'pipe');
```
ivanovSPvirtru, line 17: amazingly hasOwnProperty returns false for stream altough `!!source.pipe` returns pipe. HasOwnPropery is for innumerable props (the one that will appear in loop, native methods are not innumerable).

mustyantsev (dismissed):

lib/facade/FileClient.ts
```diff
@@ -14,7 +14,8 @@ interface FileClientConfig {
...
+  // @ts-ignore
+  return typeof source?.pipe === 'function';
```
dmihalcik-virtru, line 18:
```suggestion
  return typeof (source as NodeJS.ReadableStream)?.pipe === 'function';
```

dmihalcik-virtru (changes requested):

dmihalcik-virtru (approved):

PR #85, opened 2022-09-02
Throttle rewrap on failures

krokhmaliuk-virtru:
https://virtru.atlassian.net/browse/PLAT-1905


lib/src/index.ts
```diff
@@ -368,13 +371,26 @@ export class NanoTDFDatasetClient extends Client {
 
     const version = '0.0.1';
     // Rewrap key on every request
+    if (this.isRewrapFailed) {
+      throw new Error('Key rewrap failure');
```
pflynn-virtru, line 375:
```suggestion
      throw new Error('Key rewrap throttled');
```

lib/src/index.ts
```diff
@@ -368,13 +371,26 @@ export class NanoTDFDatasetClient extends Client {
 
     const version = '0.0.1';
     // Rewrap key on every request
+    if (this.isRewrapFailed) {
```
pflynn-virtru, line 374: Are we able to inspect `nanoTdfHeader`?  If we can add an inexpensive check to see if the Policy changed.  So throttle if the same policy issued that got a 403.  If the policy is different then hit rewrap endpoint.

lib/package.json
```diff
@@ -44,6 +44,7 @@
   "scripts": {
     "build": "npm run clean && webpack && tsc -p tsconfig-esm.json && node prepareNodeESM.js && ../scripts/add-module-types.sh esm && ../scripts/fill-version.sh",
     "clean": "rm -rf {build,dist}",
+    "updateLink": "npm run build && sudo npm rm @opentdf/client -g && sudo npm link",
```
obidenko, line 47: do we rly need it ?

obidenko (approved):

lib/src/index.ts
krokhmaliuk-virtru: Updated

lib/package.json
pflynn-virtru: how is this used?  can you add to README

dmihalcik-virtru (changes requested):

dmihalcik-virtru:
This is fine, but doesn't really reflect the best practice for how to handle token eviction. Basically, we can assume the 'auth' service is the the object responsible for the quality and accuracy of the tokens, not the client. Instead of maintaining a list of 'bad' tokens, it should evict the 'current' token from the auth object and get a new one. I'm fine with this solution for now, as it will work, but it has several problems:

- It means the calling application will have to initialize a new 'oidc-auth' object (or whole client) whenever their token expires
- even if it did work, using a list doesn't scale very well, e.g. if it works as expected on a page that we just leave open and we get a new token on the 'fails' list every hour (or even every day) the list is probably not the best

In short, we need an `evict` method on the auth class or something, instead of this. This would let the auth providers or the OIDC client track these issues and appropriately respond, e.g. by running a token refresh, or perhaps throwing an appropriate error or somehow getting the user to log back in with a redirect

dmihalcik-virtru (approved):
This is fine, but still requires manually running `updateLink` when files change. If we simplified the project to eliminate the `prepareNodeESM` and `add-module-types` helpers, we could just use the build-in `--watch` options to web pack and tsc to eliminate that need

PR #98, opened 2022-10-10
🐛 Fix broken conditional exports

dmihalcik-virtru:
- Fix types for tdf-stream aliased module
- Update types around tdfstream generally to be more sensible without needing to load the alias module

### DIFF
--- a/lib/tdf3/src/client/index.ts
+++ b/lib/tdf3/src/client/index.ts
@@ -10,7 +10,7 @@ import {
 } from '../utils/index';
 import { base64 } from '../../../src/encodings/index';
 import TDF from '../tdf';
-import { PlaintextStream } from './tdf-stream';
+import { type AnyTdfStream, TdfStream } from './tdf-stream';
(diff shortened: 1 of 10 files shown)


lib/tdf3/src/client/index.ts
```diff
@@ -226,7 +237,7 @@ export class Client {
     output,
     rcaSource = false,
     windowSize = DEFAULT_SEGMENT_SIZE,
-  }: EncryptParams) {
+  }: EncryptParams): Promise<AnyTdfStream | null> {
```
elizabethhealy, line 240: did you mean to repeat the function definition?

elizabethhealy (approved):

lib/tdf3/src/client/index.ts
dmihalcik-virtru: This is an overloaded signature.

https://howtodoinjava.com/typescript/function-overloading/

The reason for doing this here is the return type depends on the presence of the 'output' named parameter - if present, this method returns `null` ; otherwise, it creates and returns a new stream

PR #101, opened 2022-10-26
⬆️ Require node 18 for node users

dmihalcik-virtru:
- Node 18 includes a global 'fetch' command, removing the need to include `node-fetch`
- Notably, node-fetch fails when run from a CommonJS `require` block now, so this also adds a commonJS node test to prevent any regression on loading @opentdf/client in a node app that uses CJS


pflynn-virtru (approved):
since we are adding more test directories, perhaps we should move them under `tests` directory.

PR #102, opened 2022-11-01
🔥 Remove need to readable stream aliases

dmihalcik-virtru:
- WebStreams are supported in node 16.5+, so require those as needed.
- Translate a few remaining uses of non-web streams into web streams
- Remove aliases


dmihalcik-virtru:
@danielRicaud Do you know what the `@runtimeConfig` alias is/was used for?

pflynn-virtru (approved):

PR #113, opened 2022-11-28
🐛 Typescript fixes

dmihalcik-virtru:
- Disables @ts-ignore on tdf.js
- Enable strict mode in lint again
- Fixes a bunch of errors that fall out of that:
  - jose import of x509 had incorrect call
  - upsert response not being set

### DIFF
--- a/.github/workflows/format.yaml
+++ b/.github/workflows/format.yaml
@@ -13,9 +13,15 @@ jobs:
           cache-dependency-path: './lib/package-lock.json'
       - run: make format
       - name: ignore package-lock changes
-        run: git checkout -- {cli,web-app}/package-lock.json
+        run: git restore "*/package-lock.json"
+      - run: git diff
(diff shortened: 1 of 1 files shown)


danielRicaud (dismissed):

lib/tdf3/src/tdf.ts
```diff
@@ -871,10 +873,9 @@ class TDF extends EventEmitter {
       this.manifest.encryptionInformation.method.algorithm.toLowerCase()
     );
 
-    const encryptedSegmentSizeDefault =
-      parseInt(
-        this.manifest.encryptionInformation.integrityInformation.encryptedSegmentSizeDefault
-      ) || DEFAULT_SEGMENT_SIZE;
+    const defaultSegmentSize =
```
sujankota, line 876: I thin we want `DEFAULT_SEGMENT_SIZE` not 0

sujankota (approved):

PR #116, opened 2022-11-30
✨ DPoP Headers in creds

dmihalcik-virtru:


dmihalcik-virtru:
Instead of using the same signing key for tdf3, we have to generate a second one since JWT doesn't like the algorithm parameters we use. This probably could be worked around - we can re-cast the key via import/export, for example - but I doubt it is worth the hassle. See https://github.com/opentdf/client-web/pull/118

PR #117, opened 2022-12-02
🎨 Refactor auth middleware as function

dmihalcik-virtru:
- Unify on a middleware method, `withCreds`
- Previously, the OIDC middleware only could return one value, Authorization header, which means we can't also set the DPoP header
- The appID middleware updates HttpRequest in place, but that isn't best practice; instead, return a copy-and-modified version
- Related fixes to use the middeware where needed
- This pre-work simplifies adding DPoP headers in the middleware
- Adds validation that access token is present in CLI tool


lib/src/auth/auth.ts
```diff
@@ -1,4 +1,45 @@
-import { AxiosRequestConfig } from 'axios';
+import { JWTPayload, SignJWT } from 'jose';
+
+export type Method =
```
pflynn-virtru, line 3: nit: I expected to see UPPERCASE here

pflynn-virtru (approved):

lib/src/auth/auth.ts
dmihalcik-virtru: yeah the spec is very shouty, reading it now

pflynn-virtru (approved):

PR #119, opened 2022-12-12
draft of migration

ivanovSPvirtru:


lib/package.json
```diff
@@ -105,15 +106,15 @@
     "@web/test-runner-commands": "~0.6.5",
     "audit-ci": "^6.3.0",
     "babel-loader": "^8.2.5",
-    "babel-plugin-module-resolver": "^4.1.0",
```
ivanovSPvirtru, line 108: ```fix available via `npm audit fix --force`
Will install babel-plugin-module-resolver@2.2.0, which is a breaking change```
Thats for audit issue

lib/src/auth/Eas.ts
```diff
@@ -35,7 +34,7 @@ class Eas {
     endpoint: string;
     requestFunctor?: RequestFunctor;
   }) {
-    this.authProvider = authProvider.withCreds;
+    this.authProvider = authProvider;
```
ivanovSPvirtru, line 37: There were bug with context (this) cause it was executed from other object.

lib/src/auth/auth.ts
```diff
@@ -14,16 +14,24 @@ export type Method =
 /**
  * Generic HTTP request interface used by AuthProvider implementers.
  */
-export type HttpRequest = {
+export class HttpRequest {
```
ivanovSPvirtru, line 17: We use it like this so I had to make it class back

lib/tdf3/src/tdf.ts
```diff
@@ -300,7 +300,7 @@ class TDF extends EventEmitter {
       }
       encryptionInformation.keyAccess.push(keyAccess);
     }
-
+    debugger;
```
dmihalcik-virtru, line 303:
```suggestion
```

dmihalcik-virtru (approved):

dmihalcik-virtru (approved):

PR #120, opened 2022-12-22
Changes needed to integrate with rca-link-service

danielRicaud:


pflynn-virtru (approved):

lib/tdf3/src/client/builders.ts
```diff
@@ -794,8 +794,25 @@ class DecryptParamsBuilder {
   /**
    * @param rcaParams
    */
-  setRcaSource(rcaParams: RcaParams) {
-    this._params.rcaSource = rcaParams;
+  setRcaSource(rcaParams: RcaParams | RcaLink) {
```
dmihalcik-virtru, line 797: Is this a copy of the existing builder method to maintain compatibility? If this is not, I'd suggest moving the logic out into a `urlToRcaParams` function and export that, to allow for better tree shaking and more clear errors - generally I don't like setters to throw

lib/tdf3/src/client/builders.ts
```diff
@@ -794,8 +794,25 @@ class DecryptParamsBuilder {
...
+        const wk = rcaLink.get('wk'); // split knowledge
+        const al = rcaLink.get('al');
+        if (!pu || !wu || !wk || !al) {
+          throw new Error('RCA link is not valid');
+        }
+        this._params.rcaSource = { pu, wu, wk, al };
+      } catch (e) {
+        console.error('RCA link is not valid', e);
+        throw new Error('RCA link is not valid');
+      }
```
dmihalcik-virtru, line 814: Reorganize the logic so you aren't re-throwing and include more details.
```suggestion
        const al = rcaLink.get('al');
      } catch (cause) {
        throw new Error('RCA link is not valid', {cause});
      }
      if (!pu || !wu || !wk || !al) {
        throw new Error(`RCA link [${rcaParams}] is missing parameters!`);
      }
      this._params.rcaSource = { pu, wu, wk, al };
```

dmihalcik-virtru, line 814: Let's get more accurate checks and more details.

```suggestion
    try {
      let p = rcaParams;
      if (typeof rcaParams === 'string') {
        p = Object.fromEntries(new URLSearchParams(rcaParams));
      }
      const {pu, wu, wk, al} = p;
      if (!pu || !wu || !wk || !al) {
        throw new Error('RCA link is missing parameter(s)');
      }
      this._params.rcaSource = { pu, wu, wk, al };
    } catch (cause) {
      throw new Error('RCA link is not valid', {cause});
    }
```

dmihalcik-virtru (dismissed):

lib/tdf3/src/client/index.ts
```diff
@@ -261,6 +261,8 @@ export class Client {
     payloadKey,
   }: EncryptParams): Promise<AnyTdfStream | null> {
     if (rcaSource && asHtml) throw new Error('rca links should be used only with zip format');
+    if (rcaSource && !this.kasEndpoint)
```
danielRicaud, line 264: Is this a valid assumption? Since RCA links will need a policy uuid ?

lib/tdf3/src/client/index.ts
```diff
@@ -100,6 +100,7 @@ export interface ClientConfig {
   clientId?: string;
   dpopEnabled?: boolean;
   kasEndpoint?: string;
+  easEndpoint?: string;
```
pflynn-virtru, line 103: are we still using this? pooey

lib/tdf3/src/client/builders.ts
```diff
@@ -203,6 +203,10 @@ class EncryptParamsBuilder {
    * @return {EncryptParamsBuilder} - this object.
    */
   withStreamSource(readStream: ReadableStream<Uint8Array>): EncryptParamsBuilder {
+    if (!inBrowser() && !readStream?.getReader) {
+      readStream = toWebReadableStream(readStream);
+    }
+
```
dmihalcik-virtru, line 209: The type here is ReadableStream, so just assert that it is if you are worried about using this method outside of a TS context
```suggestion
    assert readStream.getReader;
```

lib/tdf3/src/client/builders.ts
```diff
@@ -700,6 +704,10 @@ class DecryptParamsBuilder {
    * @return {DecryptParamsBuilder} - this object.
    */
   withStreamSource(stream: ReadableStream<Uint8Array>) {
+    if (!inBrowser() && !stream?.getReader) {
+      stream = toWebReadableStream(stream);
+    }
+
```
dmihalcik-virtru, line 710: This is redundant with the setStreamSource method implementation
```suggestion
```

pflynn-virtru (dismissed):

lib/tdf3/src/utils/zip-writer.ts
```diff
@@ -100,7 +100,7 @@ export class ZipWriter {
   zip64: boolean;
   constructor() {
     // Enable zip64
-    this.zip64 = false;
+    this.zip64 = true;
```
dmihalcik-virtru, line 103: Can you pull the zip64 changes out into a separate PR for ease of rollback in case they don't work right?

lib/tdf3/src/client/builders.ts
```diff
@@ -1,16 +1,16 @@
 import { S3Client, GetObjectCommand, HeadObjectCommand, S3ClientConfig } from '@aws-sdk/client-s3';
 import axios from 'axios';
-
+import { toWebReadableStream } from 'web-streams-node';
```
dmihalcik-virtru, line 3: Instead of doing this, just add a note telling users to first run their node stream through `streams.Readable.toWeb()`.
```suggestion
```

dmihalcik-virtru (changes requested):

lib/tdf3/src/client/index.ts
danielRicaud: I know we're moving to OIDC, but I couldn't locally test this feature in a browser in development without being able to alter the `accounts` endpoint

lib/tdf3/src/utils/zip-writer.ts
danielRicaud: Yes let me yank those out good idea! Although Misha tested them against the c++ core it's better safe than sorry 👍

lib/tdf3/src/client/builders.ts
danielRicaud: The more built-ins we use the better! Thanks for catching this! I'll implement and test all your streams.readable.toWeb() feedback!

lib/tdf3/src/client/index.ts
dmihalcik-virtru: It's fine to have this error here, if only as a precaution

dmihalcik-virtru (dismissed):

dmihalcik-virtru:
d'oh @danielRicaud you'll need to rebase as the autoformat commit isn't signed, e.g. using the 'squash' tool in GitHub desktop or with `git rebase -i` from the command line

danielRicaud:
I think updating OSX broke my signing somehow! Thanks for the tip!

dmihalcik-virtru:
d'oh it looks like we need to update the package-locks

PR #121, opened 2022-12-23
🔓 Only warn on insecure pubkey lookup

dmihalcik-virtru:
- This allows the lookup to succeed in private environments
- Fix bug in CLI where it would return 0 process status on unknown errors
- Adds `.strict()` to yargs parser to help prevent errors due to misspelled args
- Discovered these while fixing backend's tests to use the latest beta of this library: https://github.com/opentdf/backend/pull/381


pflynn-virtru (approved):

dmihalcik-virtru:
@pflynn-virtru fixed a failing `bats` test due to a changed error name

pflynn-virtru (approved):

PR #127, opened 2023-01-10
🤖 Remove fill-version post-build script

dmihalcik-virtru:
This is breaking source maps. Instead, have the version hard coded and update the version bump  and check-version scripts to change/validate the manually input versions


dmihalcik-virtru:
An alternative is to have the `esm` output also run through webpack, which has the convenient https://webpack.js.org/plugins/define-plugin/, which the script was replacing really

ivanovSPvirtru (approved):

PR #130, opened 2023-01-11
🔥 remove webpack

dmihalcik-virtru:
- `tsc` is good enough for our needs as a library app
- Moves a lot of logic that was in webpack into tsconfig-* files
- Uses ts-loader instead of babel loader for mocha
- Tests use `esmock` instead of `proxyquire`
- some tests use rollup internally, but we don't use it for the build step


ivanovSPvirtru (approved):
Works with SR and Sec Share

PR #131, opened 2023-01-16
Locker complience fixed

ivanovSPvirtru:
Locker api also doesn't like dynamic import


lib/src/auth/AccessToken.ts
```diff
@@ -136,7 +137,6 @@ export class AccessToken {
     }
     if (this.signing_key) {
       // TODO: dpop does not have a CommonJS variant.
-      const { default: dpopFn } = await import('dpop');
```
dmihalcik-virtru, line 139: I think this is to work around the dpop library not exporting a CJS variant. Node supports dynamic import even in CJS so we may have to do something else here - or fork the library, most likely

See: https://stackoverflow.com/a/71093947

ivanovSPvirtru: Damn it. Welp, last time i had such issues with locker complience, ive just putted lib as a folder. Not whole thing but what we use. Also we could fork it and add webpack to bundle CJS variant. Or maybe theres other libs that doing what drop does

image

dmihalcik-virtru: I'm thinking we can fork as opentdf and publish to GitHub's package registry. Adding the CommonJS output should be pretty similar to what we are doing with this library now

dmihalcik-virtru (approved):

sievdokymov-virtru (approved):

PR #138, opened 2023-01-27
✨ adds tdf3 testing to web app

dmihalcik-virtru:
- Adds tdf3 encrypt (with and without html envelope) to web app, and corresponding decrypt
- Adds playwright coverage
- Uses the `Client` instead of `FileClient`
- Needed some type fixups to types to let it work without builders, e.g. adding sensible defaults requires making some parameters optional
- Since we aren't bundling, the application is responsible for adding the Buffer polyfill. I'm guessing we haven't seen this issue since we are pulling it in via the virtru-sdk bundler for our internal apps. See https://github.com/opentdf/client-web/pull/144 for a potential solution to this problem for direct users of this library one the web


dmihalcik-virtru:
I'm grabbing a draft PR so I can reference this code for changes I need in backend and quickstart to enable port forwarding through a local host web app, required to avoid CORS and general single-origin issues when using `fetch` with both keycloak and KAS.

pflynn-virtru (approved):

PR #141, opened 2023-02-01
🎨 simplify tsconfig names

dmihalcik-virtru:
- tsconfig-common can just be tsconfig
- remove duplicated (already set) fields in things that extend it
- fixes misnamed 'test' folder, actually is 'tests' (plural)
- fixes several several sonar cloud discovered issues, including a broken method


lib/tsconfig-node-esm.json
```diff
@@ -1,9 +1,8 @@
 {
-  "extends": "./tsconfig-common.json",
+  "extends": "./tsconfig.json",
   "compilerOptions": {
-    "module": "ES2020",
+    "module": "es2020",
```
pflynn-virtru, line 4: sonarcloud is recommending this to be lowercase?  after doing a scan, the internet can't decide if it should be upper or lower.

pflynn-virtru (approved):

lib/tsconfig-node-esm.json
dmihalcik-virtru: I went with lower for consistency with the typescript docs: https://www.typescriptlang.org/tsconfig#module

PR #148, opened 2023-02-13
Added rca support

danielRicaud:
…-link-service-aws-2


package.json
```diff
@@ -0,0 +1,5 @@
+{
```
dmihalcik-virtru, line 1: Delete these stray package and package-lock files from the project root

dmihalcik-virtru (commented):
looks good just remove the added files

lib/tdf3/src/client/DecoratedReadableStream.ts
```diff
@@ -33,7 +33,7 @@ export abstract class DecoratedReadableStream {
   upsertResponse?: UpsertResponse;
 
   constructor(byteLimit: number, underlyingSource: UnderlyingSource) {
```
dmihalcik-virtru, line 35: If we don't allow byteLimit to be set anymore should we go through and remove it entirely?

lib/tdf3/src/client/DecoratedReadableStream.ts
```diff
@@ -32,8 +32,8 @@ export abstract class DecoratedReadableStream {
   manifest: Manifest;
   upsertResponse?: UpsertResponse;
 
-  constructor(byteLimit: number, underlyingSource: UnderlyingSource) {
-    this.stream = new ReadableStream(underlyingSource, { highWaterMark: byteLimit });
+  constructor(underlyingSource: UnderlyingSource) {
+    this.stream = new ReadableStream(underlyingSource, { highWaterMark: 1 });
```
dmihalcik-virtru, line 36: IIUC this will always immediately write whatever to the underlying stream. Which makes sense, as this is just supposed to be a proxy, not a buffer.

dmihalcik-virtru (approved):

PR #159, opened 2023-03-10
🚧 FileSystem API in web-app

dmihalcik-virtru:
Replaces download link generation with FileSystemAccess API implementation of streaming downloads for tdf3 (and html wrapped, kinda) containers. Works fine for tdf files (not HTML wrapped) of arbitrary size.

HOWEVER: the showSaveFilePicker is unsupported by Playwright, so any CI implementation might have to mock the whole thing.

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -1147,9 +1147,9 @@
       }
     },
     "node_modules/@aws-sdk/util-locate-window": {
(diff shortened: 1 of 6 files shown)


dmihalcik-virtru:
Moved this work into the bigger demo PR https://github.com/opentdf/client-web/pull/164

PR #160, opened 2023-03-10
✅ Adds web-app remote file test

dmihalcik-virtru:
- Adds a 'url' input field to the web app
- Adds a simple node:http test server that can serve exactly one file (with range header support)
- Works on small (~100 MB) files but fails on large (GB+) files


dmihalcik-virtru:
Hah it is failing on encrypt, not decrypt, so it isn't the chunker bug. I suspect it is just not throttling the download but IDK it could be a leak somewhere? It is strange that File streams don't trigger it but web streams do. I've had no luck so far, see https://github.com/opentdf/client-web/pull/159

PR #162, opened 2023-03-14
working on decrypt rca large files

danielRicaud:


lib/tdf3/src/tdf.ts
```diff
@@ -941,60 +941,64 @@ export class TDF extends EventEmitter {
...
+        const segmentIntegrityAlgorithmType =
+          this.manifest?.encryptionInformation.integrityInformation.segmentHashAlg;
+        const segmentHashStr = await this.getSignature(
+          reconstructedKeyBinary,
+          Binary.fromBuffer(encryptedChunk),
+          segmentIntegrityAlgorithmType || integrityAlgorithmType
+        );
 
-          controller.enqueue(decryptedSegment.payload.asBuffer());
+        if (segment?.hash !== base64.encode(segmentHashStr)) {
```
dmihalcik-virtru, line 976:
```suggestion
        if (segment.hash !== base64.encode(segmentHashStr)) {
```

lib/tdf3/src/tdf.ts
```diff
@@ -941,60 +941,64 @@ export class TDF extends EventEmitter {
...
+          centralDirectory,
+          '0.payload',
+          encryptedOffset,
+          encryptedSegmentSize
+        );
+        encryptedOffset += encryptedSegmentSize;
+
+        // use the segment alg type if provided, otherwise use the root sig alg
+        const segmentIntegrityAlgorithmType =
+          this.manifest?.encryptionInformation.integrityInformation.segmentHashAlg;
```
dmihalcik-virtru, line 969: Instead of cluttering up the code with `?.` save the manifest above and assert that it is present like you do with the `segment` variable.

lib/tdf3/src/tdf.ts
```diff
@@ -941,60 +941,64 @@ export class TDF extends EventEmitter {
...
       },
-    });
+    };
+
+    const outputStream = makeStream(underlyingSource);
+
+    if (rcaParams && rcaParams.wu) {
+      let res = await axios.head(rcaParams.wu);
+
+      outputStream.fileSize = parseInt(res?.headers?.['content-length'] as string, 10) || undefined;
```
dmihalcik-virtru, line 1000: You may want to log some details here if this doesn't work, as we probably want this to be an error or warning for us if it fails, since it indicates a configuration problem with the storage server

lib/tdf3/src/utils/chunkers.ts
```diff
@@ -1,8 +1,22 @@
-import axios, { AxiosResponse } from 'axios';
+// import axios, { AxiosResponse } from 'axios';
 import { Buffer } from 'buffer';
 import { createReadStream, readFile, statSync } from 'fs';
 import { type AnyTdfStream, isAnyTdfStream } from '../client/tdf-stream.js';
 
+// const keepaliveAgent = new HttpsAgent({
```
dmihalcik-virtru, line 6: Can you delete this code (and remove any additional deps from package.json)?

lib/tdf3/src/utils/chunkers.ts
```diff
@@ -71,40 +85,42 @@ export const fromNodeFile = (filePath: string): Chunker => {
...
-      throw e;
+      }),
+    })
+    if(!res.ok) {
+      throw new Error(
+        'Unexpected response type: Server should have responded with an ArrayBuffer.'
+      );
     }
+
+    // @ts-ignore
```
dmihalcik-virtru, line 103: What are you ts-ignoring here?

lib/tests/mocha/unit/chunkers.spec.ts
```diff
@@ -267,27 +267,29 @@ describe('chunkers', () => {
...
       const { fromUrl } = await import('../../../tdf3/src/utils/chunkers.js');
-      const c: Chunker = fromUrl(urlFor(path));
+      const c: Chunker = await fromUrl(urlFor(path));
       const one: Uint8Array = await c(1, 2);
       expect(one).to.deep.equal(b.slice(1, 2));
       expect(Array.from(one)).to.deep.eq([1]);
     });
     it('negative one', async () => {
       const { fromUrl } = await import('../../../tdf3/src/utils/chunkers.js');
+      // @ts-ignore
```
dmihalcik-virtru, line 284: Please at least add comments here. You can also usually do something more specific.

Also current preferred behavior is to use `ts-expect-error` for test code, and avoid `ts-ignore` as much as possible for new code

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#ts-ignore-or-ts-expect-error

lib/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "forceConsistentCasingInFileNames": true,
     "inlineSourceMap": true,
     "isolatedModules": true,
-    "lib": ["dom", "dom.iterable", "es2020", "ES2022.Error"],
+    "lib": ["dom", "dom.iterable", "es2020", "ES2022.Error", "webworker"],
```
dmihalcik-virtru, line 13: Are we though?

dmihalcik-virtru (changes requested):

lib/tdf3/src/client/BrowserTdfSteam.ts
```diff
@@ -1,31 +1,44 @@
...
 
 export class BrowserTdfStream extends DecoratedReadableStream {
   override async toFile(filepath = 'download.tdf'): Promise<void> {
     if (isFirefox()) {
-      await fileSave(new Response(this.stream), {
-        fileName: filepath,
-        extensions: [`.${filepath.split('.').pop()}`],
-      });
-      return;
+      let ponyfill = await import('web-streams-polyfill');
```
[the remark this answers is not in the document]
danielRicaud: Yeah it's strange but Firefox was complaining until I added it saying it couldn't find a constructor.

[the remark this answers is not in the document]
danielRicaud: it's wrapped in a `if( isFirefox() ) {}` if that's enough signal to someone looking through this codebase in the future?

lib/package.json
```diff
@@ -64,7 +64,9 @@
   },
   "dependencies": {
     "@aws-sdk/client-s3": "3.197.0",
+    "agentkeepalive": "^4.3.0",
```
dmihalcik-virtru, line 67: Please remove the unused deps

lib/tdf3/src/client/BrowserTdfSteam.ts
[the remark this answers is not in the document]
dmihalcik-virtru: yeah this is fine if we need it

dmihalcik-virtru (changes requested):

lib/package.json
```diff
@@ -73,21 +73,26 @@
     "events": "^3.3.0",
     "jose": "^4.11.2",
     "jsonschema": "^1.4.1",
+    "pump": "^3.0.0",
```
dmihalcik-virtru, line 76: Can you get rid of pump and the ws-node (and their corresponding types?) Thx

lib/tdf3/src/client/NodeTdfStream.ts
```diff
@@ -4,8 +4,22 @@ import { DecoratedReadableStream } from './DecoratedReadableStream.js';
 
 export class NodeTdfStream extends DecoratedReadableStream {
   override async toFile(filepath: string, encoding: BufferEncoding = 'utf-8'): Promise<void> {
-    const nodeStream = createWriteStream(filepath, { encoding, flags: 'w' });
```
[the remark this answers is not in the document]
dmihalcik-virtru: 'web first' means do everything in web stream world where possible. Ideally we'd get rid of the node stream code from this library (@opentdf/client) entirely. We can leave it in the virtru wrapper library for compatibility

dmihalcik-virtru (dismissed):

dmihalcik-virtru (approved):

dmihalcik-virtru (approved):

PR #164, opened 2023-03-15
⚗️ Adds more test scenarios to web-app

dmihalcik-virtru:
Adds a few alternatives and buttons to see if where the different bugs might exist. notably:

1. Adds a 'how to save' selector, to allow dumping the stream, for selecting between Streamsaver (default download), fsapi (called Save As... although only Chrome actually opens the file save dialog), and 'none' (dump to a null writer). 
2. Adds a 'demo scan' button, which doesn't use our library at all but just does some text processing from a Jake Archibald demo
3. Adds a rate limited progress logger to the console
4. Adds 'cancel' button to abort longer streams

### DIFF
--- a/lib/tdf3/src/client/NodeTdfStream.ts
+++ b/lib/tdf3/src/client/NodeTdfStream.ts
@@ -12,7 +12,7 @@ export class NodeTdfStream extends DecoratedReadableStream {
   ): Promise<void> {
     const encoding =
       (typeof options === 'string' && options) || (options && options.encoding) || 'utf-8';
-    const nodeStream = createWriteStream(filepath, { encoding, flags: 'w' });
+    const nodeStream = createWriteStream(filepath, { encoding, flags: 'w', highWaterMark: 1 });
     const writer = Writable.toWeb(nodeStream);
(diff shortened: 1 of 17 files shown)


lib/tdf3/src/client/NodeTdfStream.ts
```diff
@@ -1,9 +1,17 @@
...
 
 export class NodeTdfStream extends DecoratedReadableStream {
-  override async toFile(filepath: string, encoding: BufferEncoding = 'utf-8'): Promise<void> {
+  override async toFile(
+    filepath: string,
+    options: BufferEncoding | DecoratedReadableStreamSinkOptions = 'utf-8'
+  ): Promise<void> {
+    const encoding =
+      (typeof options === 'string' && options) || (options && options.encoding) || 'utf-8';
     const nodeStream = createWriteStream(filepath, { encoding, flags: 'w' });
```
danielRicaud, line 15: add explicit high water mark

danielRicaud (approved):
😎

danielRicaud (approved):
😎

PR #165, opened 2023-03-21
Remove Virtru and Secure Reader references

jrschumacher:


dmihalcik-virtru (commented):
AFAICT we can't customize this? We'll probably have to add a set of parameters to customize these; it looks like this feature was forgotten in the move to open source

PR #172, opened 2023-04-06
Add format params to fetchEntityObject

krokhmaliuk-virtru:


sievdokymov-virtru (dismissed):

ivanovSPvirtru (dismissed):

lib/src/auth/Eas.ts
```diff
@@ -51,6 +51,9 @@ class Eas {
       url: this.endpoint,
       method: 'POST',
       headers: { 'Content-Type': 'application/json' },
+      params: {
```
dmihalcik-virtru, line 54: This will return the JWT variant in a response header, IIUC, so we should get two values - the old EO and the new JWT. We can then use the JWT for v2 endpoints and EO for v1 endpoints, either somehow deciding based on what version of KAS we have or a parameter to the client constructor.

PR #173, opened 2023-04-06
Added axios retry logic when fetching remote chunk from url

danielRicaud:


lib/tdf3/src/utils/chunkers.ts
```diff
@@ -2,6 +2,10 @@ import axios, { AxiosResponse } from 'axios';
 import { Buffer } from 'buffer';
 import { createReadStream, readFile, statSync } from 'fs';
 import { type AnyTdfStream, isAnyTdfStream } from '../client/tdf-stream.js';
+import axiosRetry from 'axios-retry';
+
+// @ts-ignore
+axiosRetry(axios, { retries: 3 }); // Retries all idempotent requests (GET, HEAD, OPTIONS, PUT, DELETE)
```
dmihalcik-virtru, line 8: Am I nuts for wanting an exponential delay as well? It sounds like it retries immediately by default. We also may want to console.log via an onRetry handler so we can get some details, unless you know that it already does that. I'm fine without it since we can get the information from the network tab, but it might be helpful if we want customers to send us logs

https://www.npmjs.com/package/axios-retry

dmihalcik-virtru (approved):
See my comment for some optional 'enhancements'

PR #176, opened 2023-04-14
Changes streamsavers service worker, and fixes vscode TS error

danielRicaud:


lib/.eslintrc.yaml
danielRicaud, line 1: Fixes the following VScode error, has anyone else seen this at the top of `.ts` files in client-web?

Screenshot 2023-04-14 at 5 08 48 PM

lib/tdf3/src/client/BrowserTdfSteam.ts
```diff
@@ -22,6 +22,8 @@ export class BrowserTdfStream extends DecoratedReadableStream {
       return;
     }
 
+    streamSaver.mitm = 'https://secure.virtru.com/rca/stream-saver/index.html';
```
danielRicaud, line 25: Any place we can un-hardcode this and make it dynamic per environment? Although Auth-Widget is hardcoded to prod so maybe this isn't a deal breaker?

pflynn-virtru: no Virtru urls in OpenTDF

danielRicaud: Any opinions on where this should live? Since stream-saver is only utilized in client-web?

pflynn-virtru (approved):

PR #179, opened 2023-04-26
Added retry on chunk download on any error

danielRicaud:


lib/tdf3/src/utils/chunkers.ts
```diff
@@ -5,7 +5,11 @@ import { type AnyTdfStream, isAnyTdfStream } from '../client/tdf-stream.js';
 import axiosRetry from 'axios-retry';
 
 // @ts-ignore
-axiosRetry(axios, { retries: 3, retryDelay: axiosRetry.exponentialDelay }); // Retries all idempotent requests (GET, HEAD, OPTIONS, PUT, DELETE)
+axiosRetry(axios, {
+  retries: 3,
+  retryDelay: axiosRetry.exponentialDelay,
+  retryCondition: () => true,
```
dmihalcik-virtru, line 11: We probably don't want to retry on every failure. Do you know what the default condition is?

Also, does this affect all global axios requests? We might want to start using `axios.create` to allow more fine-grained control - and avoid accidentally altering behavior outside of what we intend

dmihalcik-virtru: Also it looks like you can set these parameters on a given request (at least, if paired with axios.create, although IDK if that is required or just in the example code), so we maybe want to be more selective about this specific change

danielRicaud: Previously it'd only retry indempodent requests and now it also included POST/PATCH along with the indempodent methods. It also now retries timeout related errors like `ECONNABORTED` that would previously not be retried.



image

dmihalcik-virtru (approved):

PR #184, opened 2023-05-09
Initial implementation of parallelize downloads for RCA

danielRicaud:


lib/tdf3/src/tdf.ts
```diff
@@ -951,60 +952,129 @@ export class TDF extends EventEmitter {
...
 
     const underlyingSource = {
       pull: async (controller: ReadableStreamDefaultController) => {
-        if (segments.length === 0) {
-          controller.close();
-          return;
-        }
+        pullController = controller;
+        // @ts-ignore
+        if (rcaParams && rcaParams.wu) {
```
dmihalcik-virtru, line 967: This should be a new option, not recycling `.wu`. AFAICt this is independent of the chunker type / source. e.g. it should work on any chunk able source, right?

```suggestion
        if (concurrentReadsEnabled) {
```

and then add `concurrentReadsEnabled` to the method parameters

dmihalcik-virtru:
we've removed remote store support

PR #187, opened 2023-05-22
Hotfix/cks header

krokhmaliuk-virtru:


lib/tdf3/src/tdf.ts
```diff
@@ -514,6 +514,7 @@ export class TDF extends EventEmitter {
           body.clientPayloadSignature = await reqSignature(body, pkKeyLike);
         }
         const httpReq = await this.authProvider.withCreds(this.buildRequest('POST', url, body));
+        httpReq.headers['X-Virtru-Public-Key'] = btoa(this.publicKey);
```
dmihalcik-virtru, line 517: Can you add a FIXME or TODO here to clean this up? It isn't part of the spec and exists to work around a bug in some internal Virtru implementations of authProvider. 
```suggestion
        // TODO: Move this logic into internal virtru authProviders
        httpReq.headers['X-Virtru-Public-Key'] = btoa(this.publicKey);
```

dmihalcik-virtru (approved):

PR #195, opened 2023-06-08
🚨 Require allowlisting KASes on decrypt and upsert

dmihalcik-virtru:
Why? TDF files should be considered user generated content, and any data from them should be suspect. In this case, we don't want to send random upsert / rewrap requests to unknown/unsafe URIs.

This throws an UnsafeUrlError in this case, which has a `.url` parameter which will allow applications to prompt the user to see if the URL is 'safe'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 
-version=1.2.2
+version=1.3.0
 pkgs=lib web-app
 
 .PHONY: all audit license-check lint test ci i start format clean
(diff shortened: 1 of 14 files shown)


dmihalcik-virtru:
@ivanovSPvirtru LMK when the audit fix is merged thx

danielRicaud (approved):
Looks good! Thanks for adding the comments and JSDocs 😎

PR #203, opened 2023-07-09
Refactor

arkavo-com:
Working package that can be imported into webpack 5 app.  see https://github.com/arkavo-org/client-web/pkgs/npm/client


dmihalcik-virtru:
What is the goal here? Getting rid of the monorepo?

PR #204, opened 2023-07-10
✨ Enables CryptoService DI

dmihalcik-virtru:
- Adds an optional `cryptoService` parameter to the TDF3 constructor.
- Exposed the relevant types
- Updates target version to 1.1.0 since this is a new feature.


lib/tdf3/src/crypto/declarations.ts
```diff
@@ -21,12 +22,14 @@ export type PemKeyPair = {
  */
 export const MIN_ASYMMETRIC_KEY_SIZE_BITS = 2048;
 
-export type CryptoService = {
+export type CryptoService<PairT = CryptoKeyPair> = {
```
pflynn-virtru, line 25: what does this mean?

lib/tdf3/src/crypto/declarations.ts
```diff
@@ -59,13 +67,15 @@ export type CryptoService = {
    * Generate an RSA key pair
    * @param size in bits, defaults to a reasonable size for the default method
    */
-  generateKeyPair: (size?: number) => Promise<CryptoKeyPair>;
+  generateKeyPair: (size?: number) => Promise<PairT>;
```
pflynn-virtru, line 70: `PairT` means?

pflynn-virtru (approved):

lib/tdf3/src/client/index.ts
```diff
@@ -89,12 +90,13 @@ const makeChunkable = async (source: DecryptSource) => {
   }
   // Unwrap if it's html.
   // If NOT zip (html), convert/dump to buffer, unwrap, and continue.
-  const htmlBuf = buf || (await initialChunker());
+  const htmlBuf = buf ?? (await initialChunker());
```
danielRicaud, line 93: Nice catch

danielRicaud (approved):

lib/tdf3/src/crypto/declarations.ts
dmihalcik-virtru: This is to allow the types of the keys to vary with the crypto service. Notably, a custom provider may keep its keys in its own database or HSM or something, so it might not want to keep the CryptoKey/CryptoKeyPair interface

lib/tdf3/src/crypto/declarations.ts
dmihalcik-virtru: I'll add more documentation soon, but this is to be a key pair type - an abstraction of CryptoKeyPair

lib/tdf3/src/client/index.ts
dmihalcik-virtru: sonar cloud found it for me

PR #206, opened 2023-07-11
⌨️ Lets crypto generators be async

dmihalcik-virtru:
- generally, crypto primitives should be async since they may call out to external sources of entropy
- Updates type exports generally to include some more named types used by both the crypto SPI and by virtru-sdk
- A couple of small fixes and cleanups from the last PR as well


lib/tdf3/src/crypto/index.ts
```diff
@@ -69,7 +69,7 @@ export function rsaPkcs1Sha256(
  * Generate a random hex key
  * @return New key as a hex string
  */
-export function generateKey(length?: number): string {
+export async function generateKey(length?: number): Promise<string> {
```
danielRicaud, line 72: Why do some of the Promisified functions `return` and some `Promise.resolve()`? Guessing because they're synchronous but just need to "look like" a promise to the api?

dmihalcik-virtru: It is just inconsistent style. Declaring a function `async` will auto-wrap all return values with Promise.resolve so I'll do that instead.

danielRicaud: ohhhhh learned something new today!

danielRicaud (approved):

PR #220, opened 2023-07-20
🏷️ Remove generic from CryptoService

dmihalcik-virtru:
I don't understand them well enough to make a good generic interface yet


dmihalcik-virtru:
The idea: data types differ between crypto services, so we should have a way to specify them. Reality: they should be encapsulated with the same interface between implementations. For now, this means keys-pairs have two fields: publicKey and privateKey. The problem is the types of *those* values varies, and I don't want to say 'any'. What do I put instead? I think this conjunction type is the best option for now

danielRicaud (approved):

PR #222, opened 2023-07-22
🏷️ Expose WebCryptoService to importers

dmihalcik-virtru:


dmihalcik-virtru:
Explicit is better

lib/tdf3/index.ts
```diff
@@ -14,6 +14,7 @@ import {
   type EncryptResult,
   type PemKeyPair,
 } from './src/crypto/declarations.js';
+export * as WebCryptoService from './src/crypto/index.js';
```
biscoe916, line 17: Nit, but I'd prefer all the exports being on the bottom of the file. 

I can make the change.

dmihalcik-virtru: done

danielRicaud (approved):

PR #223, opened 2023-07-31
Feature/without buffer proto

ivanovSPvirtru:


lib/tdf3/src/ciphers/aes-gcm-cipher.ts
```diff
@@ -46,14 +46,14 @@ export class AesGcmCipher extends SymmetricCipher {
    * it's parts.  There is no need to process the payload.
    */
   override async encrypt(payload: Binary, key: Binary, iv: Binary): Promise<EncryptResult> {
-    const toConcat: Buffer[] = [];
+    const toConcat: Uint8Array[] = [];
     const result = await this.cryptoService.encrypt(payload, key, iv, Algorithms.AES_256_GCM);
-    toConcat.push(iv.asBuffer());
-    toConcat.push(result.payload.asBuffer());
+    toConcat.push(new Uint8Array(iv.asArrayBuffer()));
```
cassandrabailey293, line 51: how often does `encrypt` get called? will it result in any memory issues?

lib/tdf3/src/utils/buffer-crc32.ts
```diff
@@ -0,0 +1,46 @@
...
+  0x95bf4a82, 0xe2b87a14, 0x7bb12bae, 0x0cb61b38, 0x92d28e9b, 0xe5d5be0d, 0x7cdcefb7, 0x0bdbdf21,
+  0x86d3d2d4, 0xf1d4e242, 0x68ddb3f8, 0x1fda836e, 0x81be16cd, 0xf6b9265b, 0x6fb077e1, 0x18b74777,
+  0x88085ae6, 0xff0f6a70, 0x66063bca, 0x11010b5c, 0x8f659eff, 0xf862ae69, 0x616bffd3, 0x166ccf45,
+  0xa00ae278, 0xd70dd2ee, 0x4e048354, 0x3903b3c2, 0xa7672661, 0xd06016f7, 0x4969474d, 0x3e6e77db,
+  0xaed16a4a, 0xd9d65adc, 0x40df0b66, 0x37d83bf0, 0xa9bcae53, 0xdebb9ec5, 0x47b2cf7f, 0x30b5ffe9,
+  0xbdbdf21c, 0xcabac28a, 0x53b39330, 0x24b4a3a6, 0xbad03605, 0xcdd70693, 0x54de5729, 0x23d967bf,
+  0xb3667a2e, 0xc4614ab8, 0x5d681b02, 0x2a6f2b94, 0xb40bbe37, 0xc30c8ea1, 0x5a05df1b, 0x2d02ef8d,
+]);
+
+export function unsigned(buf: Uint8Array, previous: number) {
```
cassandrabailey293, line 38: o0o0o0o bitwise operators, neato - looks like this is also from that link. i guess we can't use directly because it's not in TS? could potentially contribute a TS pr back to that repo, but it looks like he hasn't touched it in a year or so

lib/tdf3/src/utils/zip-reader.ts
```diff
@@ -1,6 +1,6 @@
-import { Buffer } from 'buffer';
 import { Manifest } from '../models/index.js';
 import { Chunker } from './chunkers.js';
+import { readUInt32LE, readUInt16LE, copyUint8Arr } from './index.js';
```
cassandrabailey293, line 3: you can say `index.js` instead of `index.ts` here?

cassandrabailey293 (commented):
Looks good to me! Any docs or tests need to be updated?

lib/tdf3/src/ciphers/aes-gcm-cipher.ts
ivanovSPvirtru: Ill test that to be sure.

lib/tdf3/src/utils/buffer-crc32.ts
dmihalcik-virtru: I think maybe because they use node Buffers? 

Types are provided via the 'DefinitivelyTyped' crowdsourced utility project. One of two OSS contributions I've done with my work account was this

https://github.com/DefinitelyTyped/DefinitelyTyped/commits/308aae57497dd57f2568602d843e395aaa405096/types/buffer-crc32/index.d.ts

The problem is that it [requires Buffer](https://github.com/brianloveswords/buffer-crc32/blob/master/index.js#L1)

ivanovSPvirtru: We could if they will start transition. Because initial bundle configuration of this lib to setup is a not a small effort.

lib/tdf3/src/utils/zip-reader.ts
ivanovSPvirtru: Its like a bundler thing, file should be called js even when they are ts. I know, looks not very intuitive ))

dmihalcik-virtru (approved):

PR #229, opened 2023-08-04
🐛 Fix 'cause' parameter to tdf error

dmihalcik-virtru:
This hopefully will fix some missing stack traces in the console


dmihalcik-virtru:
wait I am now double wrapping it :-/

PR #235, opened 2023-09-06
fix spec char issue

ivanovSPvirtru:


lib/tdf3/src/binary.ts
```diff
@@ -60,7 +62,12 @@ function adjustSliceParams(length: number, start: number, end?: number): [number
   if (end && end < 0) {
     end = length + end;
   }
-  return [start, end];
+  const result = [start];
+  if (end) {
+    result.push(end);
+  }
```
ivanovSPvirtru, line 68: Because [number, number | undefined] is not the same with [number, number?]

pflynn-virtru (dismissed):

PR #240, opened 2023-10-01
😬 Add chrome origin trials to template

dmihalcik-virtru:
- This allows template users to opt-out of top-frame storage partitioning, for one


khvirtru:
> * This allows template users to opt-out of top-frame storage partitioning, for one

this is a temporary workaround, and will have an actual fix so that it can work with Storage Partitioning, correct?

PR #241, opened 2023-10-03
test added more

ivanovSPvirtru:


lib/tdf3/src/client/index.ts
```diff
@@ -400,13 +404,16 @@ export class Client {
       metadata,
     });
 
+    const { keyForEncryption, keyForManifest, keyForLink } = await distributeKeysMiddleware(!!rcaSource, payloadKey)
```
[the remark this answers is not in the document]
ivanovSPvirtru: Great ideas. We can get it squeaky clean. And i dont even thing we need to one middleware return enother. Its just make sense to make two middlewares. KeyMiddlware and StreamMiddlware. Maybe you want to do something with stream before you return it? here is simple way to do it. And in virtru-sdk we can do whatever.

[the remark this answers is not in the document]
ivanovSPvirtru: Updated also this
https://github.com/virtru/virtru-tdf3-js/pull/586

lib/tdf3/src/client/builders.ts
```diff
@@ -2,10 +2,15 @@ import { AttributeValidator } from './validation.js';
 import { AttributeObject, Policy } from '../models/index.js';
 import { type RcaParams, type RcaLink, type Metadata } from '../tdf.js';
```
dmihalcik-virtru, line 3: Can you move these RCA types and references to virtru-sdk as well? This would include moving some methods and fields from the builders here to their subclasses in virtru-sdk

I think we can do this without requiring a major (or even minor?) version update to virtru-sdk.

lib/tdf3/src/client/index.ts
```diff
@@ -330,7 +331,8 @@ export class Client {
     offline,
     windowSize,
     eo,
-    payloadKey,
+    keyMiddleware = keyMiddleware,
+    streamMiddleware = stream => stream,
```
dmihalcik-virtru, line 335: Nice this is a more straightforward than my suggestion.

lib/tdf3/src/utils/index.ts
```diff
@@ -278,3 +282,28 @@ function base64clean(str: string) {
...
+}>{
+  const tdfService = TDF.create({ cryptoService: WebCryptoService });
+  tdfService.setEncryption({ type: 'split' });
+  if (!tdfService.encryptionInformation?.generateKey) {
+    throw new Error('Crypto service not initialised')
+  }
+
+  const key1 = await tdfService.encryptionInformation.generateKey();
+
+  return { keyForEncryption: key1, keyForManifest: key1 };
```
dmihalcik-virtru, line 308: Isn't the default using the same key for both?

dmihalcik-virtru (commented):
going in the right direction, esp with regards to more code moving into virtru-sdk

lib/tdf3/src/utils/index.ts
ivanovSPvirtru: Yep but its default middleware is here
https://github.com/virtru/virtru-tdf3-js/pull/586/files#diff-5631231d382b7c1a4ac85b4c2a941793890a0f5f12348a232a76e2fd3b45ae89R99

and set as default middlware. https://github.com/opentdf/client-web/pull/241/files#diff-549436fb06c43a51e130655ea4292b8cedd3a77dd8364603423f4fe3cc5b3865R360

Or Virtru-sdk has two usecases, CKS and RCA which case its always different key

lib/tdf3/src/client/index.ts
```diff
@@ -395,17 +387,18 @@ export class Client {
       metadata,
     });
 
+    const { keyForEncryption, keyForManifest } = await keyMiddleware();
```
dmihalcik-virtru, line 390: I feel like we need a context parameter here? IDK if it isn't needed for virtru-sdk maybe we should leave it out. The idea would be a user could use the same keyMiddleware object across multiple encrypt calls or something? 

I am fine with leaving it blank for now as adding a new thing is easier than changing or removing something

dmihalcik-virtru (changes requested):
Please remove the macOS `.DS_Store` files

dmihalcik-virtru (approved):

PR #243, opened 2023-10-09
✨ KAS Key Identifier support

dmihalcik-virtru:
- Sends `v=2` when requesting kas_public_key for tdf3 content
- When connecting to a KAS that implements the new draft v2 return format, this will result in a response with the kid field present
- This is then baked into the key access object for future reference

### DIFF
--- a/lib/tdf3/src/tdf.ts
+++ b/lib/tdf3/src/tdf.ts
@@ -42,7 +42,9 @@ import {
   ManifestIntegrityError,
   PolicyIntegrityError,
   TdfDecryptError,
+  TdfError,
   TdfPayloadExtractionError,
+  UnsafeUrlError,
 } from '../../src/errors.js';
 import { htmlWrapperTemplate } from './templates/index.js';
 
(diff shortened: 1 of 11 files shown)


lib/tdf3/src/tdf.ts
```diff
@@ -125,6 +128,97 @@ type TDFConfiguration = {
...
+    };
+  } catch (cause) {
+    if (cause?.response?.status != 400) {
+      throw new TdfError(
+        `Retrieving KAS public key [${kas}] failed [${cause.name}] [${cause.message}]`,
+        cause
+      );
+    }
+  }
+  // Retry with v1 params
```
pflynn-virtru, line 199: it will get here if any exception is thrown unless the exception has a response code of 400?

dmihalcik-virtru:
I had added some additional strictness checks to the KAS url, but found they didn't work with the current test environment we have that uses simple HTTP to tunnel. My hope is that the check in the Client constructor (and Builder) will be enough, without the new check I had added to the fetchKasPublicKey function itself

pflynn-virtru (approved):

PR #244, opened 2023-10-17
Feature/download optimised

ivanovSPvirtru:


lib/src/auth/providers.ts
```diff
@@ -134,7 +134,6 @@ export const clientAuthProvider = async (clientConfig: OIDCCredentials): Promise
 };
 
 export * from './auth.js';
-export * from './OIDCCredentials.js';
```
dmihalcik-virtru, line 137: Can you also delete this file? It is empty

lib/tdf3/src/tdf.ts
```diff
@@ -1107,6 +1089,47 @@ export class TDF extends EventEmitter {
...
+          (decryptedChunk) => {
+            if (!decryptedChunk) return;
+            slice[index].decryptedChunk = decryptedChunk;
+            if (slice[index]._resolve) {
+              (slice[index]._resolve as (value: unknown) => void)(null);
+            }
+          }
+        )
+      );
+      if (promises.length === batchSize) {
```
dmihalcik-virtru, line 1126: There is a bug here - it won't wait for all batches if slice.length isn't a multiple of batchSize.

Suggestion: Use `p-limit` instead: https://www.npmjs.com/package/p-limit

This has the benefit of not batching, so may even be faster, and should simplify this code overall

dmihalcik-virtru (commented):
A couple of small suggestions.

dmihalcik-virtru (approved):

PR #256, opened 2023-11-16
Feature/workers unoptimised

ivanovSPvirtru:


lib/tdf3/src/tdf.ts
```diff
@@ -981,6 +982,10 @@ export async function sliceAndDecrypt({
   cryptoService: CryptoService;
   segmentIntegrityAlgorithm: IntegrityAlgorithm;
 }) {
+  const limit = pLimit(navigator.hardwareConcurrency || 4); // save fallback number
```
dmihalcik-virtru, line 985: We might want to make this customizable by the caller somehow? (See also: chunksInOneDownload)

lib/tdf3/src/crypto/decrypt-worker.ts
```diff
@@ -0,0 +1,75 @@
...
+      self.postMessage({ success: true, data: decryptedData });
+    } catch (error) {
+      self.postMessage({ success: false, error: error.message });
+    }
+  };
+`;
+
+const workerBlob = new Blob([workerScript], { type: 'application/javascript' });
+const workerUrl = URL.createObjectURL(workerBlob);
+const workersArray: Worker[] = Array.from({ length: maxWorkers }, () => new Worker(workerUrl));
```
dmihalcik-virtru, line 26: Ideally we'd be doing this at build time, not runtime. Do we have any methods for doing this? I can't find a good way to do this without doing a two-pass build.

lib/tdf3/src/crypto/decrypt-worker.ts
```diff
@@ -0,0 +1,75 @@
...
+
+interface WorkersQueue {
+  freeWorkers: Worker[];
+  resolvers: ((worker: Worker) => void)[];
+  push: (worker: Worker) => void;
+  pop: () => Promise<Worker>;
+}
+
+const workersQueue: WorkersQueue = {
```
dmihalcik-virtru, line 35: Can you add some test coverage for this?

lib/tdf3/src/crypto/decrypt-worker.ts
ivanovSPvirtru: This code gets executate in the instance when we import entry point of sdk to any frontend app. Because this file eventually gets imported from entry point further down the nodes. And imports are synchronous. ANd code above too.

lib/tdf3/src/tdf.ts
ivanovSPvirtru: I dont think we do tbh. Ive benchmarked it tried multiple configurations. More plimit with hardwareConcurrency amount of workers, vice-verca, more both of them. What i got: 
1)  number of Workers more than amount of cores really didnt bosted performance
2) pLimit ether didnt had any effect ether were so big that gained small improvement but froze ui

lib/tdf3/src/crypto/decrypt-worker.ts
ivanovSPvirtru: Oh yes i tried it and i had constant problems on a app side. Ive tuned it on Virtru-sdk side but after webpack on secure share it had constant errors. Spend a wile on that. 

Thats why i decided to go with this aproach. Its makes sdk independent of bundlers issues and doesnt slow down performance, since its a singleton instance that initializes workers as soon as sdk imported it. So when youll start download stuff its already be there.

PR #257, opened 2023-12-04
INT-790: Remove extra await

ntrevino-virtru:
This is to fix large file downloads from what I understand from @ivanovSPvirtru 😅

### DIFF
--- a/lib/tdf3/src/tdf.ts
+++ b/lib/tdf3/src/tdf.ts
@@ -947,7 +947,7 @@ async function updateChunkQueue(
             bufferSize
           );
           if (buffer) {
-            await sliceAndDecrypt({
+            sliceAndDecrypt({
               buffer,
               cryptoService,
               reconstructedKeyBinary,
--- a/lib/tdf3/src/tdf.ts
+++ b/lib/tdf3/src/tdf.ts
@@ -999,7 +999,7 @@ export async function sliceAndDecrypt({
       segmentIntegrityAlgorithm,
       cryptoService
     )
-      .then(result => {
+      .then((result) => {
         slice[index].decryptedChunk = result;
         return null;
       })
--- a/lib/tdf3/src/tdf.ts
+++ b/lib/tdf3/src/tdf.ts
@@ -111,6 +111,7 @@ type Chunk = {
   encryptedOffset: number;
   encryptedSegmentSize?: number;
   decryptedChunk?: null | DecryptResult;
+  promise: Promise<unknown>;
   _resolve?: (value: unknown) => void;
   _reject?: (value: unknown) => void;
 };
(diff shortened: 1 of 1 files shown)


lib/tdf3/src/tdf.ts
```diff
@@ -1061,7 +1061,8 @@ export async function readStream(cfg: DecryptConfiguration) {
 
   const cipher = new AesGcmCipher(cfg.cryptoService);
 
-  await updateChunkQueue(
```
dmihalcik-virtru, line 1064: Hmm. This means if there is a network error in this function, it won't propagate anymore. Instead of throwing, you will have to explicitly call a _reject handler on the chunk.

- In this method: add a corresponding `_reject` parameter, in addition to the `_resolve` handler below. This then propagate an error in chunk processing to the stream assembly code here by throwing it out on the await on line 1085
```
      if (!chunk.decryptedChunk) {
        await new Promise((resolve, reject) => {
          chunk._resolve = resolve;
          chunk._reject = reject;
        });
      }
```
- Above, in `updateChunkQueue`, instead of throwing the error, send it to the offending slice. It is kinda hard, since you won't know which slice the error happened on in the current double-looped code. So you'll have to refactor it a bit so the handler happens within the innermost loop, and add another handler for the zip format check which just attaches it to the first item in the stride.

dmihalcik-virtru (changes requested):

lib/tdf3/src/tdf.ts
ntrevino-virtru: I think it should be hooked up properly, though not sure exactly which zip format check you meant.

lib/tdf3/src/tdf.ts
```diff
@@ -981,27 +982,33 @@ export async function sliceAndDecrypt({
...
 }) {
-  for (const index in slice) {
-    const { encryptedOffset, encryptedSegmentSize } = slice[index];
-
-    const offset =
-      slice[0].encryptedOffset === 0 ? encryptedOffset : encryptedOffset % slice[0].encryptedOffset;
-    const encryptedChunk = new Uint8Array(
-      buffer.slice(offset, offset + (encryptedSegmentSize as number))
-    );
+  return Promise.all(
```
ivanovSPvirtru, line 985: Unfortunetly it wont change anything. When you do Promise.All on decrypt its not paralelise decryption. I tried it donesnt give any peformnace boost.

https://github.com/opentdf/client-web/pull/256/files#diff-5c6d9ea24cc7a3470e79efacc4f7aed59b657798a7cba23499d0cfd6476b9935R970

here work that im doing for that uses workers. But without limiting promise.all of lot of chunks can freeze ui so i use pLimit.

Could you remove `Promise.all` for now since changes is on the way? and leave just _reject part?

ntrevino-virtru: Done.

lib/tdf3/src/tdf.ts
```diff
@@ -946,7 +947,7 @@ async function updateChunkQueue(
             bufferSize
           );
           if (buffer) {
-            sliceAndDecrypt({
+            await sliceAndDecrypt({
```
ivanovSPvirtru, line 950: We dont need await here. Dowload (getPayloadSegment) doesnt need to wait for dowloaded chunk to decrypt to start to dowload next chunk. Its just slowedown file download. 
image

stream waits for new chunk in chun map, slice and decrypt provide it. Dowload doesnt have to stop.

ntrevino-virtru: Thanks for the explanation! Updated!

ivanovSPvirtru (approved):

ntrevino-virtru:
There was still an issue with decrypts, but I was able to test in SS and it seems fine to me so far.

dmihalcik-virtru (approved):
You could rewrite it to not use .then, but the logic seems correct as is if you'd rather not

PR #264, opened 2024-04-12
chore: Replace backend with platform

dmihalcik-virtru:
We have migrated to a new set of backend services and are deprecating the existing python based services. This work in progress replaces the sample web application code and all of the code that brings up a backend and configures some users.


dmihalcik-virtru:
Replacing this PR with https://github.com/opentdf/client-web/pull/282

Issue #273, opened 2024-05-08
Services SDK generation and implementation


cassandrabailey293:
Add SDK support for calling services so web clients can easily access data like a list of attributes or fetch data like the attributes which the current user should have access to.

Tool to generate code: https://github.com/bufbuild/protobuf-es
Example of build with registry and usage: https://github.com/connectrpc/examples-es/tree/main/react/vite

## Acceptance Criteria

- create a workflow that can be executed manually or automated to build the generated code
- use the upstream protos (opentdf/platform) in a dynamic way so that we can build protos from a supported git tag (aka version)
- add an example to the README to explain how someone might access the service SDKs
	- focus on connection and not on a specific example
- (optional) expose the service SDK in a convenient way to improve developer experience -- this work could be split out as a separate issue

PR #280, opened 2024-05-09
feat(lib): generate ts code from platform protos

suchak1:


jrschumacher:
@dmihalcik-virtru @pbacon-blaber the requirements I gave Krish are #273. Testing with the platform was not a requirement and we will work on this in the coming days/week with @ryanulit as he is the first developer needing this in web.

pbacon-blaber (approved):
LGTM especially regarding @jrschumacher comment saying it's expected that this isn't tested and will be tested / tweaked if necessary when @ryanulit uses it for the first time

PR #282, opened 2024-05-13
chore(ci): Adds platform integration test

dmihalcik-virtru:
We have migrated to a new set of backend services and are deprecating the existing python based services. This replaces the recommended sample code with platform compatible code


dmihalcik-virtru:
Starting a new PR as my old one is hard to merge after some time

pflynn-virtru (approved):

PR #286, opened 2024-05-16
feat: Adds back cli tool

dmihalcik-virtru:
We [removed this CLI sample code last summer](https://github.com/opentdf/client-web/pull/207) as part of a move to go all-in on web-first code. There are still some bits of 'node-style' code around - notably we still use EventEmitter and Buffer (provided via ponyfills) - but largely have moved away from other bits of node io or networking apis. As such we have to migrate our code that previously used node's files and streams to the (newer) webstreams support.

While restoring this code, I made some changes to simplify calling the client from node:

- Simplifies ability to set attributes in TDF3Encrypt (no longer require attribute objects with KAS url)
- Removes ability to pass in an output stream to TDF3Client.encrypt

In order to work around streamsaver failing in node contexts, the cli also adds a small polyfill to get that to at least only report an error.

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -1,0 +1,422 @@
+import './polyfills.js';
+import { openAsBlob } from 'node:fs';
+import { open, readFile, stat, writeFile } from 'node:fs/promises';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
+import {
+  type AuthProvider,
+  type EncryptParams,
+  type HttpRequest,
+  AuthProviders,
+  NanoTDFClient,
+  NanoTDFDatasetClient,
+  TDF3Client,
+  version,
+  EncryptParamsBuilder,
+  DecryptParams,
+  DecryptParamsBuilder,
+} from '@opentdf/client';
+import { CLIError, Level, log } from './logger.js';
+import { webcrypto } from 'crypto';
+
+type AuthToProcess = {
+  auth?: string;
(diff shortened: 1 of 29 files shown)


dmihalcik-virtru:
This builds on 
1. https://github.com/opentdf/client-web/pull/270
2. https://github.com/opentdf/client-web/pull/285
3. https://github.com/opentdf/client-web/pull/284

cli/src/cli.ts
```diff
@@ -0,0 +1,422 @@
...
+            log('DEBUG', `TDF3 Client`);
+            const client = new TDF3Client({ authProvider, kasEndpoint, dpopEnabled: argv.dpop });
+            log('SILLY', `Initialized client ${JSON.stringify(client)}`);
+            const ct = await client.encrypt(await tdf3EncryptParamsFor(argv));
+            if (!ct) {
+              throw new CLIError('CRITICAL', 'Encrypt configuration error: No output?');
+            }
+            if (argv.output) {
+              const o = createWriteStream(argv.output);
+              await ct.stream.pipeTo(o);
```
pflynn-virtru, line 360: I changed this to 
```
import * as fs from 'fs/promises';
...
              const filehandle = await fs.open(argv.output, 'w');
              await filehandle.writeFile(ct.stream);
```

dmihalcik-virtru: ok

pflynn-virtru (approved):

PR #299, opened 2024-06-03
fix(dpop): respect dpop disabled flag in oidc access token class methods

jakedoublev:
Resolves https://github.com/opentdf/client-web/issues/300


dmihalcik-virtru (approved):

dmihalcik-virtru:
yes, it can be up to KAS to enforce a requirement on the signature

PR #302, opened 2024-06-07
chore(docs): Cleanup readmes

dmihalcik-virtru:
- Remove references to opentdf/opentdf and backend repos
- rename tdf3 to ztdf in at least one place
- clarify that nodejs is *not* a supported runtime
- update howto run the sample web app to reflect what we are doing in ci


jakedoublev (approved):

web-app/tests/README.md
```diff
@@ -1,52 +1,64 @@
...
-tilt up
+```sh
+go run github.com/opentdf/platform/service@latest provision keycloak
+./config-demo-idp.sh
 ```
 
-Finally, bring up the web app:
+If you haven't already done so, initialize your service keys.
+Be careful!
+If you lose these keys you will lose access to all TDFs encrypted with them.
```
patmantru, line 29: These keys are stored in the directory `<where>`

dmihalcik-virtru: For now they are just created whereever you put them. They are loaded by o/platform/service from the current working directory.

patmantru (approved):

PR #307, opened 2024-07-15
feat(sdk): Allow custom KAO array templates

dmihalcik-virtru:
This will allow opening of files with multi-kas splits.

Step 1 of #314

While I'm here I also:

1. Lets platform-xtest job run in parallel with other e2e tests
2. Updates config for platform-roundtrip job, since the `provision fixture keycloak` needs a new config file
3. Adds a new phony make target to simplify development, you can now `make cli` to get the cli without having to guess the version number
4. Adds a `prettier ignore` directive so `make format` no longer formats the generated protocol buffer code
5. Updates the fetching of KAS public keys to happen during encrypt, not during client creation. This makes more sense now that we don't know at client start time which kases will actually be involved in the encrypt step


pflynn-virtru (dismissed):

pflynn-virtru (dismissed):

lib/tdf3/src/client/index.ts
```diff
@@ -384,14 +384,23 @@ export class Client {
...
-        url: kasPublicKey.url,
-        kid: kasPublicKey.kid,
-        publicKey: kasPublicKey.publicKey,
-        metadata,
+
+    const splits: SplitStep[] = splitPlan || [{ kas: this.kasEndpoint }];
+    encryptionInformation.keyAccess = await Promise.all(
+      splits.map(async ({ kas, sid }) => {
+        if (!(kas in this.kasKeys)) {
+          this.kasKeys[kas] = fetchKasPublicKey(kas);
```
elizabethhealy, line 392: does the cache have a time limit similar to the go one?

elizabethhealy (approved):

lib/tdf3/src/client/index.ts
dmihalcik-virtru: no it probably should?

dmihalcik-virtru: however this is an existing feature; I'm just changing it to be lazily loaded while before it was loaded during the client constructor

dmihalcik-virtru: https://github.com/opentdf/client-web/issues/326

PR #313, opened 2024-07-25
fix(ci): Remove old backend-roundtrip ci step

elizabethhealy:


dmihalcik-virtru:
Are we intending to continue to support secure share with this code base? It is still based off of backend (underneath some layers of proprietary addons)

PR #315, opened 2024-07-31
fix(deps): Updates deps

dmihalcik-virtru:
- unable to update `eslint` and `@aws` deps, as they have breaking changes
- otherwise everything else is updated, including typescript, which required a few minor changes


dmihalcik-virtru:
This appears to be breaking some stuff so I may hold off on this for a bit

PR #321, opened 2024-08-08
fix(client): Normalize allowlist to origins

dmihalcik-virtru:


elizabethhealy (commented):
is there a place where were adding back the /kas before hitting rewrap? that looks like the issue with the tdfviewer error, the kasurl from the manifest doesnt have the /kas?
also could this cause issues if you have multiple kases hosted at the same origin?

pflynn-virtru (approved):

dmihalcik-virtru:
> is there a place where were adding back the /kas before hitting rewrap? that looks like the issue with the tdfviewer error, the kasurl from the manifest isnt actually pointing to the kas? could there be issues if you have multiple kases hosted at the same origin?

The intention of the allowlist is really to prevent cross origin siphoning of credentials, so I think matching on origin is sufficient

lib/tdf3/src/client/index.ts
```diff
@@ -271,9 +271,10 @@ export class Client {
       this.kasEndpoint = clientConfig.keyRewrapEndpoint.replace(/\/rewrap$/, '');
     }
 
+    const kasOrigin = new URL(this.kasEndpoint).origin;
     if (clientConfig.allowedKases) {
-      this.allowedKases = [...clientConfig.allowedKases];
-      if (!validateSecureUrl(this.kasEndpoint) && !this.allowedKases.includes(this.kasEndpoint)) {
+      this.allowedKases = clientConfig.allowedKases.map((a) => new URL(a).origin);
```
jakedoublev, line 276: Do we care at this time about the k8s issue where services might create TDFs pointing to the internal cluster URL (on http or with an internal port) that a browser might be unable to resolve or that might be different from outside the cluster? Is it crucial to have this validation in the frontend?

PR #333, opened 2024-08-25
fix(nano): Fixes `url` property when with kid

dmihalcik-virtru:
- Adds several tests for ResourceLocator
- Since the unit tests I've added use the hex encoding for ease of reading, this change also updates the hex encoding functions and greatly increases test coverage for them


dmihalcik-virtru:
merged with https://github.com/opentdf/client-web/pull/334

PR #336, opened 2024-08-27
fix: identifier type extraction from protocol

pflynn-virtru:
Removed the bit masking operation on the protocol variable to directly use its value for determining the identifier type. This simplification potentially fixes any errors induced by incorrect bit manipulation of the protocol field.


dmihalcik-virtru:
huh the reason it currently works is because we are setting the this.protocol to be the combined value with the identifier included. This is one of several surprising things which I've removed in #338

PR #338, opened 2024-08-28
fix(nano): Allow padding of kids

dmihalcik-virtru:
- Allow kids to be any size less than 33 bytes
- refactor: replaces ResourceLocator.parse and constructor with factory methods with more sensible ones that call the same simple value-only constructor, instead of having to double-parse
- removes some stray console.logs I was using for debugging that i let get checked in
- fixes issue where sometimes queries sent to the server would have double slashes
- Adds more error checking and some tests for the error conditions (unsupported protocol and identifier type, undersized buffers)
- Deletes some duplicated methods

### DIFF
--- a/lib/src/nanotdf/models/ResourceLocator.ts
+++ b/lib/src/nanotdf/models/ResourceLocator.ts
@@ -17,13 +17,6 @@ import ResourceLocatorIdentifierEnum from '../enum/ResourceLocatorIdentifierEnum
  * @link https://github.com/virtru/nanotdf/blob/master/spec/index.md#341-resource-locator
  */
 export default class ResourceLocator {
-  readonly protocol: ProtocolEnum;
-  readonly lengthOfBody: number;
-  readonly body: string;
-  readonly identifier: string;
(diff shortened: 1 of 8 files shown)


lib/src/nanotdf/models/ResourceLocator.ts
```diff
@@ -163,42 +168,40 @@ export default class ResourceLocator {
...
-
   /**
    * Get Identifier
    *
    * Returns the identifier of the ResourceLocator or an empty string if no identifier is present.
    * @returns { string } Identifier of the resource locator.
    */
-  getIdentifier(): string {
-    return this.identifier || '';
+  get kid(): string {
```
pflynn-virtru, line 204: Identifier is recommended  since it is used for lookups of KAS key, Remote Policy, Policy key

see https://github.com/opentdf/spec/blob/60171bc24301bf696462140222f972e697a8ae38/schema/nanotdf/README.md#341-resource-locator

pflynn-virtru (approved):

PR #339, opened 2024-08-29
fix: sequential requests instead of promise.all

biscoe916:


elizabethhealy (approved):

lib/tdf3/src/tdf.ts
```diff
@@ -868,79 +868,80 @@ async function unwrapKey({
...
-        throw new UnsafeUrlError(
-          `Unreconstructable key - no valid KAS found for split ${JSON.stringify(splitId)}`,
-          ''
-        );
-      }
-      // If we have multiple ways of getting a value, try the 'best' way
-      // or maybe retry across all potential ways? Currently, just tries them all
-      const [keySplitInfo] = Object.values(potentials);
-      const url = `${keySplitInfo.url}/${isAppIdProvider ? '' : 'v2/'}rewrap`;
+  const rewrappedKeys = [];
```
elizabethhealy, line 871: nit: `const rewrappedKeys: Uint8Array[] = [];`
but looks like the rest of the consts arent declaring types either so prolly fine as is

biscoe916: No I'll fix it. Thanks

biscoe916: Fixed, but yea, there aren't really proper types on a lot of this, and I don't want to address in this PR.

elizabethhealy (approved):

PR #349, opened 2024-09-10
feat(lib): Adds abac based KAO configuration

dmihalcik-virtru:
- Adds a new parameter to the encrypt method, `attributeValues`, which takes the output of the `GetAttributeValuesByFqns` gRPC call as its input.
- This will be followed by a more complete solution which also performs the RPC call if required: https://github.com/opentdf/client-web/pull/351


lib/src/policy/granter.ts
```diff
@@ -0,0 +1,158 @@
...
+
+type BooleanOperator = BooleanClause['op'];
+
+type ComplexBooleanClause = {
+  op: BooleanOperator;
+  children: BooleanClause[];
+};
+
+export function plan(dataAttrs: Value[]): KeyAccessTemplate[] {
+  // kases by value
```
pflynn-virtru, line 38: nit upper case KAS

pflynn-virtru (dismissed):
code looks fine.  needs a split key expert to review

lib/src/policy/granter.ts
```diff
@@ -0,0 +1,180 @@
+import { Attribute, AttributeRuleType, KeyAccessServer, Value } from './attributes.js';
+
+export type KeyAccessTemplate = {
```
elizabethhealy, line 3: nit: maybe rename to keysplitstep to keep in line with other sdks https://github.com/opentdf/platform/blob/d9a07f84ab5686fe13af82435af3201042dd7228/sdk/granter.go#L31

elizabethhealy (dismissed):

lib/src/policy/granter.ts
dmihalcik-virtru: Done

elizabethhealy (approved):

PR #351, opened 2024-09-16
feat(lib): Load abac config from policy service

dmihalcik-virtru:
- New parameters for tdf/client:
- New arguments to cli `encrypt`
   - `encrypt --autoconfigure` enables attribute lookup during encrypt and corresponding updates to the KAO
   - `--policyEndpoint` allows KAS and policy service to be hosted separately. if not set, it is inferred by removing `/kas` off of the end of the `--kasEndpoint` argument.
- New cli command, `attrs`, which prints out the JSON hydrated version of the attributes from the policy service
- I've broken out the ABAC configuration logic into a separate PR, https://github.com/opentdf/client-web/pull/349, to simplify review. This PR is focused on the policy service and CLI additions


lib/src/policy/attributes.ts
```diff
@@ -89,6 +89,10 @@ export type Attribute = {
   metadata?: Metadata;
 };
 
+// This is not currently needed by the client, but may be returned.
+// Setting it to unknown to allow it to be ignored for now.
+export type SubjectMapping = unknown;
```
pflynn-virtru, line 94: @jakedoublev is there an option to not return Subject Mappings ?

jakedoublev: I think [they are always returned](https://github.com/opentdf/platform/blob/b7d694d5df3867f278007660c32acb72c868735e/service/policy/db/query.sql#L146) after the migration in policy to `sqlc`. Mind filing an issue if that is a behavior that is a concern?

pflynn-virtru:
Does this policy endpoint require admin privileges?

pflynn-virtru (approved):

dmihalcik-virtru:
> Does this policy endpoint require admin privileges?

In the default casbin it is [marked as 'read' for standard access](https://github.com/opentdf/platform/blob/b7d694d5df3867f278007660c32acb72c868735e/service/internal/auth/casbin.go#L93), but not available for [public](https://github.com/opentdf/platform/blob/main/service/internal/auth/authn.go#L45) (unauthenticated) clients

lib/src/policy/attributes.ts
dmihalcik-virtru: Filed https://github.com/opentdf/platform/issues/1559

PR #352, opened 2024-09-17
feat(cli): Adds `--policyBinding ecdsa` option

dmihalcik-virtru:
Only effective with `-t nano`, this enables the ECDSA binding type to verify the policy is attached to the key is as expected.

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -31,7 +31,9 @@ type LoggedAuthProvider = AuthProvider & {
   requestLog: HttpRequest[];
 };
 
-const containerTypes = ['tdf3', 'nano', 'dataset'] as const;
+const bindingTypes = ['ecdsa', 'gmac'];
+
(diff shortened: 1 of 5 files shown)


cli/src/cli.ts
```diff
@@ -219,48 +221,52 @@ export const handleArgs = (args: string[]) => {
...
           desc: 'Mime type for the plain text file (only supported for ztdf)',
           type: 'string',
           default: '',
         },
+        userId: {
+          group: 'Encrypt Options',
+          type: 'string',
+          description: 'Owner email address',
+        },
+        usersWithAccess: {
```
elizabethhealy, line 262: is this like dissem?

elizabethhealy (approved):

cli/src/cli.ts
dmihalcik-virtru: yes

PR #362, opened 2024-10-14
feat(lib): Updated error types

dmihalcik-virtru:
- Untyped `Error` objects indicate a likely bug in the library itself
  - The messages should be prefixed `internal: `
  - Samples include bad resource management or missing values in fields that should be const.
- `TdfError` should be the root for all errors an application might theoretically screen for to find out if something is wrong in their application that might be caused by TDF or this library. Includes a novel `code` field to allow tracking based on a unique(?) error code
- `ConfigurationError` should be able to be fixed by updating the application code.
- `InvalidFileError` indicates that a file is likely tampered with or corrupt, although for some errors this may also indicate something is wrong with the user KAS. There are several subtypes when there may be changes to the configuration or user settings that could potentially fix the issue.
   - `DecryptError` may indicate that the key is incorrect; this could be caused by using a remote or CKS key is out of date, indicating a failure on the server side, but at the moment we have not implemented this.
   - `IntegrityError` indicates that the segment or global hash is incorrect. This could indicate a file was generated with a deprecated library that uses a different hash calculation.
   - `UnsafeUrlError` indicates that one or more required Key Access Objects refers to a remote KAS that is not in the allowlist. You can manually check the URL and add it to the allowlist in the client constructor if is a supported KAS.
- `NetworkError` indicates a network connectivity error (e.g. during rewrap or key lookup), or a 5xx error on a service
- `UnauthenticatedError` indicates that the Bearer token or a required DPoP was not attached to a request. This is often fixable with a mix of IdP/OAuth configuration changes and changes to the application or by adding custom middleware or some combination of all these.
- `PermissionDeniedError` indicates that a service (rewrap or public key) has denied access, either due to traditional login (bearer token insufficient scope is one possibility) or due to ABAC (client entity does not have sufficient attributes for policy)
- `UnsupportedFeatureError` indicates that an enum in the file or a KAS requirement is not met, e.g. KAS uses an unsupported EC curve, or the TDF file embeds such a curve; could indicate the file was generated with a newer, experimental, or deprecated/removed feature.


lib/src/access.ts
```diff
@@ -32,22 +33,36 @@ export async function fetchWrappedKey(
...
+    });
 
-  return response.json();
+    if (!response.ok) {
+      switch (response.status) {
+        case 401:
+          throw new UnauthenticatedError(`401 for [${req.url}]`);
+        case 403:
+          throw new PermissionDeniedError(`403 for [${req.url}]`);
+        default:
```
elizabethhealy, line 55: can we catch 400 too for policy binding errors

lib/src/errors.ts
```diff
@@ -27,91 +31,84 @@ export class TdfError extends Error {
...
+ * Indicates a decrypt failure, either due to an incorrect key, corrupt ciphertext, or inappropriate key parameters.
+ */
+export class DecryptError extends InvalidFileError {
+  override name = 'DecryptError';
 }
 
-export class InvalidEphemeralKeyError extends TdfError {
-  override name = 'InvalidEphemeralKeyError';
+export class IntegrityError extends InvalidFileError {
+  override name = 'IntegrityError';
```
elizabethhealy, line 67: is this the equivalent of tamper?

elizabethhealy (dismissed):

lib/src/access.ts
dmihalcik-virtru: ah yes will do

lib/src/errors.ts
dmihalcik-virtru: This is for either `ErrSegSigValidation` or `ErrRootSigValidation` in your PR. `InvalidFileError` is equivalent to `ErrTampered`.

elizabethhealy (dismissed):

lib/tdf3/src/models/encryption-information.ts
```diff
@@ -102,7 +102,9 @@ export class SplitKey {
           : typeof metadata === 'string'
           ? metadata
           : () => {
-              throw new Error();
+              throw new Error(
```
pflynn-virtru, line 105: base error?

lib/tdf3/src/models/encryption-information.ts
```diff
@@ -137,9 +139,10 @@ export class SplitKey {
   }
 
   async write(policy: Policy, keyInfo: KeyInfo): Promise<EncryptionInformation> {
-    const algorithm = this.cipher.name;
+    const algorithm = this.cipher?.name;
     if (!algorithm) {
-      throw new Error('Uninitialized cipher type');
+      // Hard coded as part of the cipher object. This should not be reachable.
+      throw new Error('internal: uninitialized cipher type');
```
pflynn-virtru, line 145: base error?

pflynn-virtru (approved):

lib/tdf3/src/models/encryption-information.ts
dmihalcik-virtru: Yes this maybe should be invalidFileError, now that I look closer

dmihalcik-virtru: oh wait nevermind this is during `write`. Configuration error probably

lib/tdf3/src/models/encryption-information.ts
dmihalcik-virtru: While I think this is most likely an internal error, it could conceivably also be a configuration error if the client is highly customized (a la secure share)

pflynn-virtru (approved):

PR #363, opened 2024-10-16
chore(ci): Add initial CodeQL workflow configuration

jentfoo:
I want to explore using CodeQL to assist as part of our static analysis strategy. Because this tool is available for free to open source projects, opentdf is a good place to start this testing.

This PR adds an initial configuration which will scan the codebase once a week. These results can be ignored for the time being. Instead the security team will review the results and make sure we tune away any noise first (or help in opening PR's to address any valid issues found).

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,42 @@
+name: "CodeQL"
+
+on:
+  schedule:
+    - cron: '0 13 * * 1' # At 1:00 PM UTC every Monday
+  pull_request:
+    paths:
+      - '.github/workflows/codeql.yaml'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,42 @@
+name: "CodeQL"
+
+on:
+  schedule:
+    - cron: '0 13 * * 1' # At 1:00 PM UTC every Monday
+  pull_request:
+    paths:
+      - '.github/workflows/codeql.yaml'
```
dmihalcik-virtru, line 8: Do we want this to run on every PR?

dmihalcik-virtru (approved):

.github/workflows/codeql.yaml
jentfoo: In the future that is the goal.  However right now I don't want people to burden themselves with these alerts until I am able to triage and make sure that we don't have a lot of false positive noise first.

But if we decide CodeQL is a good option going forward, I would like to have the feedback provided early, at the PR.

PR #366, opened 2024-10-22
fix(core): Removes more node buffer stuff

dmihalcik-virtru:


dmihalcik-virtru:
merged with https://github.com/opentdf/client-web/pull/371

PR #370, opened 2024-10-28
feat(cli): adds `--assertions` option

dmihalcik-virtru:


dmihalcik-virtru:
merged with https://github.com/opentdf/client-web/pull/371

PR #376, opened 2024-11-06
fix: Changes to make regex patterns more efficient, accurate and simpler

jentfoo:
This PR contains changes to address regex issues discovered by CodeQL:
* https://github.com/opentdf/web-sdk/security/code-scanning/8
* https://github.com/opentdf/web-sdk/security/code-scanning/9

Changes (see also commit messages):
* First commit contains two changes.  First to ensure that `.` characters in the url and ip addresses match only the literal `.` vs interpreted as "any character".  This is both a fix to make the patterns more precise, as well as more efficent.
* The first commit also fixes the WWW_HOST TLD part of the pattern.  Specifically `^s` was switched to `a-z`.  `^s` will match any character except `s`, it is assumed that `^\s` was expected (to match any non-whitespace character).  However the `a-z` check is simpler and more precise.
* The second commit addresses potential backtrack inefficiency by replacing `[^>]*` with `[^>]*?`.  The additional `?` will result in a non-greedy match (resulting in as minimal backtracking as possible to find a match).

After discussing with @dmihalcik-virtru we decided to further simplify `WWW_HOST` pattern.  We are able to remove the `www` and negative `www` check, remove the group capturing, and simplify the subdomain pattern.  Overall this notably improves the performance and understanding of this pattern.

### DIFF
--- a/lib/tdf3/src/client/validation.ts
+++ b/lib/tdf3/src/client/validation.ts
@@ -12,10 +12,10 @@ const SCHEME = '(https?://)';
 const HOST_PORT = '([a-z0-9][a-z0-9]{1,}:[0-9]{1,4})';
 
 // validate url host be like `www.example.com`
-const WWW_HOST = '((?:www.|(?!www))([a-z0-9][a-z0-9-]*[a-z0-9].)+[^s]{2,})';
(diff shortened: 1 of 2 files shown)


dmihalcik (approved):

dmihalcik-virtru (approved):

dmihalcik-virtru:
sorry approved in wrong profile

PR #379, opened 2024-11-12
fix: keySplitInfo Promise.any Promise.all

pflynn-virtru:
Parallelize the KAS key rewrap attempts within unwrapKey for improved efficiency. 

OR Promise.any
All potential KAS sources are now processed concurrently, and the method uses the first successful result, enhancing performance and robustness. 

AND Promise.all

Errors are collected and the most relevant one is thrown if all attempts fail.

`"target": "es2021",`

DSP-287


lib/tdf3/src/tdf.ts
```diff
@@ -926,111 +927,137 @@ async function unwrapKey({
...
     }
 
-    // If we have multiple ways of getting a value, try the 'best' way
-    // or maybe retry across all potential ways? Currently, just tries them all
-    const [keySplitInfo] = Object.values(potentials);
-    const url = `${keySplitInfo.url}/${isAppIdProvider ? '' : 'v2/'}rewrap`;
+    // Try all potential KAS sources for this split
+    const errors: Array<Error | AxiosError> = [];
+    const kasPromises = Object.values(potentials).map(async (keySplitInfo) => {
+      try {
```
dmihalcik-virtru, line 944: [Promise.any](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any) does all this for you, even having a custom type `AggregateError` that is what you expect.

```suggestion
    const kasPromises = Object.values(potentials).map(async (keySplitInfo) => {
```

lib/tdf3/src/tdf.ts
```diff
@@ -926,111 +927,137 @@ async function unwrapKey({
...
+        return {
+          success: true,
+          key: new Uint8Array(decryptedKeyBinary.asByteArray()),
+          metadata,
+        };
+      } catch (e) {
+        // Store the error but don't throw yet - we might have other valid sources
+        errors.push(e);
+        return { success: false, error: e };
+      }
```
dmihalcik-virtru, line 1001:
```suggestion
```

lib/tdf3/src/tdf.ts
```diff
@@ -926,111 +927,137 @@ async function unwrapKey({
...
     });
 
-    const jwtPayload = { requestBody: requestBodyStr };
-    const signedRequestToken = await reqSignature(
-      isAppIdProvider ? {} : jwtPayload,
-      dpopKeys.privateKey
-    );
+    // Wait for all KAS attempts for this split to complete
+    const results = await Promise.all(kasPromises);
```
dmihalcik-virtru, line 1005:
```suggestion
    let error: AggregateError;
    try {
      return await Promise.any(kasPromises);
    } catch (e) {
      // maybe need to do some type checking here?
      error = e;
    }
```

dmihalcik-virtru (dismissed):
You can simplify this a little by using Promises.any, but this code should work as is.

lib/tdf3/src/tdf.ts
```diff
@@ -926,111 +926,127 @@ async function unwrapKey({
...
       }
-      throw new InvalidFileError(
-        `Unable to decrypt the response from KAS: [${e.name}: ${e.message}], response: [${e?.response?.body}]`,
-        e
-      );
+      throw error;
     }
+  });
+
+  // Race to get the first successful split result
```
elizabethhealy, line 1031: ya for this you need a successful result for each split ID

elizabethhealy (approved):

lib/tdf3/src/tdf.ts
```diff
@@ -980,58 +965,122 @@ async function unwrapKey({
...
-        } else if (e.response.status === 403) {
-          throw new PermissionDeniedError('rewrap failure', e);
-        } else if (e.response.status === 401) {
-          throw new UnauthenticatedError('rewrap auth failure', e);
-        } else if (e.response.status === 400) {
-          throw new InvalidFileError(
-            'rewrap bad request; could indicate an invalid policy binding or a configuration error',
-            e
+
+      const reconstructedKey = keyMerge([result.key]);
```
dmihalcik-virtru, line 1006: Merging one key just returns the first value
```suggestion
      const reconstructedKey = result.key;
```

lib/tdf3/src/tdf.ts
```diff
@@ -980,58 +965,122 @@ async function unwrapKey({
...
+      key: new Uint8Array(decryptedKeyBinary.asByteArray()),
+      metadata,
+    };
+  }
+
+  // Get unique split IDs to determine if we have an OR or AND condition
+  const splitIds = new Set(Object.keys(splitPotentials));
+
+  // If we have only one split ID, it's an OR condition
+  if (splitIds.size === 1) {
```
dmihalcik-virtru, line 989: This should be the same as the case where there are multiple splitIds, this branch should not be required (maybe as an optimization to avoid having a singleton all promise but I doubt that makes a big difference).

Do you have a link to the failing test this was written to address?

dmihalcik-virtru (approved):

lib/tdf3/src/tdf.ts
pflynn-virtru: split id tests are in xtest.  test_autoconfigure_double_kas, test_autoconfigure_two_kas

some more have been added and are pending.
no unit tests for this; couldn't find anything to build upon

lib/tdf3/src/tdf.ts
pflynn-virtru: I'll follow up with this in https://github.com/opentdf/web-sdk/issues/382

PR #385, opened 2024-11-14
test: checkout before role changes

elizabethhealy:
ignore the first few commits, i was checking out wrong


elizabethhealy:
just for testing

PR #388, opened 2024-11-15
fix(sdk): Disable concurrency on rewrap

dmihalcik-virtru:
- Adds new `concurrencyLimit` decrypt param, which sets a thread pool (kinda)
- Defaults value to 1

### DIFF
--- a/lib/tdf3/src/tdf.ts
+++ b/lib/tdf3/src/tdf.ts
@@ -991,7 +991,10 @@ async function unwrapKey({
     };
   }
 
-  const poolSize = concurrencyLimit === undefined ? 1 : concurrencyLimit > 1 ? concurrencyLimit : 1;
+  let poolSize = 1;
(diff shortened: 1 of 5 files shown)


lib/tdf3/src/tdf.ts
```diff
@@ -982,77 +991,44 @@ async function unwrapKey({
     };
   }
 
-  // Get unique split IDs to determine if we have an OR or AND condition
-  const splitIds = new Set(Object.keys(splitPotentials));
-
-  // If we have only one split ID, it's an OR condition
-  if (splitIds.size === 1) {
-    const [splitId] = splitIds;
+  const poolSize = concurrencyLimit === undefined ? 1 : concurrencyLimit > 1 ? concurrencyLimit : 1;
```
biscoe916, line 994: Can we break this down so it's more readable? Nested ternary's are too hard to parse, in my opinion.

dmihalcik-virtru: better?

PR #390, opened 2024-11-18
fix(sdk): Let pools take callables.

dmihalcik-virtru:
Lets the pool start the promises itself, instead of taking in promises that have already been scheduled on the task queue (!)

### DIFF
--- a/lib/tests/mocha/unit/concurrency.spec.ts
+++ b/lib/tests/mocha/unit/concurrency.spec.ts
@@ -1,23 +1,27 @@
 import { allPool, anyPool } from '../../../src/concurrency.js';
 import { expect } from 'chai';
 
+async function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
 describe('concurrency', () => {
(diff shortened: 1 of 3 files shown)


lib/tests/mocha/unit/concurrency.spec.ts
```diff
@@ -29,29 +33,46 @@ describe('concurrency', () => {
...
+            return 1;
+          },
+          b: async () => {
+            started.add('b');
+            await sleep(50);
+            return 2;
+          },
+          c: async () => {
+            started.add('c');
+            await sleep(1500);
```
cassandrabailey293, line 50: how did you pick these values? for the sleeps?

cassandrabailey293 (approved):

lib/tests/mocha/unit/concurrency.spec.ts
dmihalcik-virtru: They need to be different enough so they are probably almost never going to flake in CI. I probably am being too sensitive on the assertion, actually. I probably should somehow randomize or round-robin the order and do it several times, if I want to be safe

PR #393, opened 2024-11-25
chore!(sdk): remove v1 support

dmihalcik-virtru:
- Removes axios and XHR support - all requests now use fetch
- Removes `v1` path support for rewrap and public key
- Removes `entityObject`
- Removes AppIdAuthProvider
- Removes `upsert`


lib/src/nanotdf/Client.ts
```diff
@@ -128,8 +128,23 @@ export default class Client {
     ephemeralKeyPair?: CryptoKeyPair,
     dpopEnabled = false
   ) {
+    const enwrapAuthProvider = (a: AuthProvider): AuthProvider => {
```
pflynn-virtru, line 131: enwrap - is this a crypto term?

pflynn-virtru (approved):

lib/src/nanotdf/Client.ts
dmihalcik-virtru: I guess I should have called it proxied or delegated/delegating. I must just have wrapping on the mind

🎁

PR #397, opened 2024-12-06
feat(sdk): remove hex encoding for segment hash

sujankota:


lib/tdf3/src/models/manifest.ts
```diff
@@ -6,4 +6,5 @@ export type Manifest = {
   payload: Payload;
   encryptionInformation: EncryptionInformation;
   assertions: Assertion[];
+  tdf_spec_version: string;
```
mkleene, line 9: it looks like this is importing [this file](https://github.com/opentdf/web-sdk/blob/main/lib/tdf3/src/version.ts) which specifies version `0.2.0` of the TDF spec. Is that what we should be doing in java?

dmihalcik-virtru: No, this should somehow be the spec version, which AFAIK is 4.3.0

https://github.com/opentdf/spec/blob/3c5adf29fe9c327b2f85b89e8beee09f0b91d8c7/VERSION

lib/tdf3/src/tdf.ts
```diff
@@ -269,26 +271,35 @@ async function _generateManifest(
     // generate the manifest first, then insert integrity information into it
     encryptionInformation: encryptionInformationStr,
     assertions: assertions,
+    tdf_spec_version: version,
```
dmihalcik-virtru, line 274:
```suggestion
    tdf_spec_version: '4.3.0',
```

lib/tdf3/src/version.ts
```diff
@@ -1,2 +1,2 @@
-export const version = '0.2.0';
+export const version = '4.3.0';
```
dmihalcik-virtru, line 1: The `version` field is set automatically by the `bump-version.sh` script, so rename this to either tdfSpecVersion or targetFileFormatVersion or similar

```suggestion
export const version = '0.2.0';
export const tdfSpecVersion = '4.3.0';
```

lib/tdf3/src/tdf.ts
dmihalcik-virtru, line 274:
```suggestion
    tdf_spec_version: tdfSpecVersion,
```

lib/tdf3/src/tdf.ts
```diff
@@ -4,6 +4,8 @@ import { DecoratedReadableStream } from './client/DecoratedReadableStream.js';
 import { fetchKasPubKey as fetchKasPubKeyV2, fetchWrappedKey } from '../../src/access.js';
 import { DecryptParams } from './client/builders.js';
 import { AssertionConfig, AssertionKey, AssertionVerificationKeys } from './assertions.js';
+import { version } from './version.js';
```
dmihalcik-virtru, line 7:
```suggestion
import { tdfSpecVersion } from './version.js';
```

dmihalcik-virtru (changes requested):

lib/tdf3/src/version.ts
dmihalcik-virtru: I've deleted this file to consolidate on lib/src/version.ts

dmihalcik-virtru (approved):

dmihalcik-virtru (dismissed):

dmihalcik-virtru (approved):

PR #406, opened 2024-12-12
chore!(sdk): Remove html wrapper creation

dmihalcik-virtru:
- still allow reading html wrapped files, for transitional support


lib/tdf3/src/client/index.ts
```diff
@@ -352,7 +350,7 @@ export class Client {
     scope = { attributes: [], dissem: [] },
     autoconfigure,
     source,
-    asHtml = false,
```
elizabethhealy, line 355: why remove the false default?

elizabethhealy (approved):

lib/tdf3/src/client/index.ts
dmihalcik-virtru: We no longer need to assert it is `=== false` as that is the default (and I think has been for some time)

dmihalcik-virtru (commented):
PTAL as I had to resolve a merge conflict

elizabethhealy (approved):

PR #412, opened 2024-12-18
fix: Assertion verification key input

elizabethhealy:


.github/workflows/build.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
   web-app:
     needs:
       - lib
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
dmihalcik-virtru, line 94: Ugh. I'm guessing we are waiting for playwright? Maybe add a note here?
```suggestion
    runs-on: ubuntu-22.04 # NOTE: Playwright not yet supported on 24.04
```

.github/workflows/roundtrip/wait-and-test.sh
```diff
@@ -29,7 +29,7 @@ VITE_TDF_CFG='{"oidc":{"host":"http://localhost:65432/auth/realms/opentdf","clie
 export VITE_PROXY
 export VITE_TDF_CFG
 
-# VITE_PROXY='{"/api":"http://localhost:5432","/auth":"http://localhost:5432"}' VITE_TDF_CFG='{"oidc":{"host":"http://localhost:65432/auth/realms/tdf","clientId":"browsertest"},"kas":"http://localhost:65432/api/kas","reader":"https://secure.virtru.com/start?htmlProtocol=1"}' npm run dev
+# VITE_PROXY='{"/api":"http://localhost:5432","/auth":"http://localhost:5432"}' VITE_TDF_CFG='{"oidc":{"host":"http://localhost:65432/auth/realms/opentdf","clientId":"browsertest"},"kas":"http://localhost:65432/api/kas","reader":"https://secure.virtru.com/start?htmlProtocol=1"}' npm run dev
```
dmihalcik-virtru, line 32: Also change the `return $?` on line 161 below to `exit 2` perhaps with a comment `echo ERROR encypt/decrypt failure` or similar

dmihalcik-virtru (changes requested):

dmihalcik-virtru (approved):

PR #449, opened 2025-02-26
feat(sdk): Updates to jose 6.x

dmihalcik-virtru:
- Jose 6.x drops support for node key objects, instead requiring/using CryptoKeys natively
- This means it removes support for node 18.x, but we have already dropped support for it
- While here, updates other deps, including ones dependabot has been suggesting


jentfoo (approved):
🙌 🚀

pflynn-virtru (dismissed):

elizabethhealy (dismissed):

elizabethhealy (approved):

PR #520, opened 2025-04-03
fix(sdk):  support hex encoding signature hash for legacy sdk

eugenioenko:
**Motivation**
Hex encoding signature was removed here https://github.com/opentdf/web-sdk/pull/397 

For **decryption** `isLegacyTDF` flag was used during decryption to determine if `base64` or `base64+hex` encoding of signatures was required for the hash.
For **encryption** only `base64` was used.

Because encryption is done only with `base64` signature hashes this produces an incompatibility with legacy clients that expect the signature hash to be `base64+hex`.

Moving forward, we do not want to use `base64+hex` but we need to allow current clients to be able to encrypt files that legacy files can decrypt. This is where the motivation for this PR changes come from.

**PR Changes**
- Add `targetSpecVersion` to  `EncryptConfiguration` to force legacy encryption 
- Add `targetSpecVersion` to `generateManifest` function to allow forcing legacy tdf spec
- Add `getSignatureHash422` function to return the correct hash format for legacy mode

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -302,6 +302,9 @@ async function parseCreateZTDFOptions(argv: Partial<mainArgs>): Promise<CreateZT
       throw new CLIError('CRITICAL', 'Invalid mimeType format');
     }
   }
+  if (argv.tdfSpecVersion) {
+    c.tdfSpecVersion = argv.tdfSpecVersion as never;
+  }
   log('DEBUG', `CreateZTDFOptions: ${JSON.stringify(c)}`);
   return c;
 }
@@ -506,6 +509,12 @@ export const handleArgs = (args: string[]) => {
           group: 'Encrypt Options:',
           type: 'string',
           description: 'Owner email address',
+        },
+        tdfSpecVersion: {
+          group: 'Encrypt Options:',
+          type: 'string',
+          description: 'TDF spec version for file creation',
+          default: tdfSpecVersion,
         },
       })
 
--- a/lib/src/opentdf.ts
+++ b/lib/src/opentdf.ts
@@ -127,6 +127,9 @@ export type CreateZTDFOptions = CreateOptions & {
 
   // Preferred algorithm to use for Key Access Objects.
   wrappingKeyAlgorithm?: KasPublicKeyAlgorithm;
+
+  // TDF spec version to target
+  tdfSpecVersion?: '4.2.2' | '4.3.0';
 };
 
 // Settings for decrypting any variety of TDF file.
@@ -366,6 +369,7 @@ export class OpenTDF {
       splitPlan: opts.splitPlan,
       windowSize: opts.windowSize,
       wrappingKeyAlgorithm: opts.wrappingKeyAlgorithm,
+      tdfSpecVersion: opts.tdfSpecVersion ?? '4.3.0',
     });
     const stream: DecoratedStream = oldStream.stream;
     stream.manifest = Promise.resolve(oldStream.manifest);
--- a/lib/tdf3/src/assertions.ts
+++ b/lib/tdf3/src/assertions.ts
@@ -155,9 +155,13 @@ export async function verify(
 /**
  * Creates an Assertion object with the specified properties.
  */
+/**
+ * Creates an Assertion object with the specified properties.
+ */
 export async function CreateAssertion(
-  aggregateHash: Uint8Array,
-  assertionConfig: AssertionConfig
+  aggregateHash: Uint8Array | string,
+  assertionConfig: AssertionConfig,
+  tdfSpecVersion?: string
 ): Promise<Assertion> {
   if (!assertionConfig.signingKey) {
     throw new ConfigurationError('Assertion signing key is required');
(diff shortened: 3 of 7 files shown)


lib/tdf3/src/tdf.ts
```diff
@@ -145,6 +145,7 @@ export type EncryptConfiguration = {
   keyForEncryption: KeyInfo;
   keyForManifest: KeyInfo;
   assertionConfigs?: AssertionConfig[];
+  targetSpecVersion?: string;
```
eugenioenko, line 148: Let's see if I'm understanding this correctly. 

When hex encoding was removed in this pr it was relying on the spec version to be `3.0.0`
https://github.com/opentdf/web-sdk/pull/397

Here how it was hardcoded
https://github.com/opentdf/web-sdk/blob/6ca50e6f3b2d945c98993190784a78e960438d8f/lib/tdf3/src/tdf.ts#L261

And updated  to use schema version and not tdf version 
https://github.com/opentdf/web-sdk/pull/414

eugenioenko: So I am wondering, if the flag  `isLegacyTDF` correct.
https://github.com/opentdf/web-sdk/blob/a240d66de226f0420486d25c0c904ce2b22bac63/lib/tdf3/src/tdf.ts#L964

Is it reliable  that only tdf with manifest schema as `3.0.0` have been encoded with hex signature hash?

(Depending on how the release was cut, we might have encrypted tdf manifest with and without the hex on the version `3.0.0`)

lib/tdf3/src/tdf.ts
```diff
@@ -335,6 +336,16 @@ async function getSignature(
...
 }
 
+function getSignatureHash(signature: Uint8Array, isLegacyTDF: boolean): string {
+  return isLegacyTDF
+    ? base64.encode(hex.encodeArrayBuffer(signature)) //  <= 3.0.0 Legacy format
+    : base64.encodeArrayBuffer(signature); // >= 4.0.0 Current format
+}
+
+function isTargetSpecLegacyTDF(specVersion?: string): boolean {
+  return !!specVersion?.startsWith('3.');
```
eugenioenko, line 346: The `startsWith('3.')` is based on the current definition of the const `isLegacyTDF`
https://github.com/opentdf/web-sdk/blob/a240d66de226f0420486d25c0c904ce2b22bac63/lib/tdf3/src/tdf.ts#L964

lib/tdf3/src/tdf.ts
```diff
@@ -961,7 +979,7 @@ export async function decryptStreamFrom(
 
   // check if the TDF is a legacy TDF
   const specVersion = manifest.schemaVersion || manifest.tdf_spec_version;
-  const isLegacyTDF = !specVersion || specVersion.startsWith('3.');
+  const isLegacyTDF = !specVersion || semver.lt(specVersion, HEX_SEMVER_THRESHOLD);
```
eugenioenko, line 981: Updated this to use `4.3.0` version as cutoff.

lib/tdf3/src/tdf.ts
eugenioenko: I updated the flag to use  `semver.lt(specVersion, '4.3.0`) as cutout

lib/tdf3/src/tdf.ts
```diff
@@ -961,7 +979,7 @@ export async function decryptStreamFrom(
 
   // check if the TDF is a legacy TDF
   const specVersion = manifest.schemaVersion || manifest.tdf_spec_version;
-  const isLegacyTDF = !specVersion || specVersion.startsWith('3.');
+  const isLegacyTDF = !specVersion || compareVersions(specVersion, HEX_SEMVER_THRESHOLD) === -1;
```
eugenioenko, line 981: Updated this to use 4.3.0 version as cutoff. It was using 3.x.x previously.

lib/package.json
```diff
@@ -73,6 +73,7 @@
   },
   "dependencies": {
     "buffer-crc32": "^1.0.0",
+    "compare-versions": "^6.1.1",
```
eugenioenko, line 76: I tried adding "semver" initially but it doesn't play nicely with modules in some test cases
```
import { lt as semverLt } from 'semver';
🚧 Browser logs:
SyntaxError: The requested module './../../node_modules/semver/index.js' does not provide an export named 'lt'

```

lib/src/opentdf.ts
```diff
@@ -127,6 +127,9 @@ export type CreateZTDFOptions = CreateOptions & {
 
   // Preferred algorithm to use for Key Access Objects.
   wrappingKeyAlgorithm?: KasPublicKeyAlgorithm;
+
+  // TDF spec version to target
+  tdfSpecVersion?: '4.2.2' | '4.3.0';
```
eugenioenko, line 132: create a type for this options

lib/tdf3/src/assertions.ts
```diff
@@ -174,11 +178,29 @@ export async function CreateAssertion(
   };
 
   const assertionHash = await hash(a);
-  const combinedHash = concatenateUint8Arrays(
-    aggregateHash,
-    new Uint8Array(hex.decodeArrayBuffer(assertionHash))
-  );
-  const encodedHash = base64.encodeArrayBuffer(combinedHash);
+  let encodedHash: string;
+  switch (tdfSpecVersion || '4.3.0') {
```
eugenioenko, line 182: tdfSpecVersion  probably shouldn't be optional

lib/tdf3/src/tdf.ts
```diff
@@ -335,6 +340,30 @@ async function getSignature(
   }
 }
 
+async function getSignatureVersion422(
```
eugenioenko, line 343: Verify if there is a way to do this without versioning the function

lib/tdf3/src/tdf.ts
```diff
@@ -960,8 +1018,8 @@ export async function decryptStreamFrom(
   const encryptedSegmentSizeDefault = defaultSegmentSize || DEFAULT_SEGMENT_SIZE;
 
   // check if the TDF is a legacy TDF
-  const specVersion = manifest.schemaVersion || manifest.tdf_spec_version;
-  const isLegacyTDF = !specVersion || specVersion.startsWith('3.');
+  const specVersion = manifest.schemaVersion || manifest.tdf_spec_version || '4.2.2';
```
eugenioenko, line 1021: the `|| '4.2.2` looks out of place, it will be possible this is undefined. Make sure isTargetSpecLegacy can handle undefined spec

eugenioenko: Resolved, `4.2.2` was changed to `tdfSpecVersion` constant defined in the SDK

cli/src/cli.ts
```diff
@@ -302,6 +302,9 @@ async function parseCreateZTDFOptions(argv: Partial<mainArgs>): Promise<CreateZT
       throw new CLIError('CRITICAL', 'Invalid mimeType format');
     }
   }
+  if (argv.tdfSpecVersion) {
+    c.tdfSpecVersion = argv.tdfSpecVersion as never;
```
dmihalcik-virtru, line 306: I think I used a type guard in my branch? You should probably do that to give reasonable error messages here, instead of blind casting

dmihalcik-virtru (approved):
While fine as is, you could improve this by adding a new TdfTargetVersion type and a corresponding type guard to filter strings or unknowns into it.

cli/src/cli.ts
eugenioenko: This is a great catch. I will update

PR #530, opened 2025-04-08
fix(sdk): Fixes generating ztdf 4.2.2 output

dmihalcik-virtru:


lib/tdf3/src/tdf.ts
```diff
@@ -401,7 +401,7 @@ export async function writeStream(cfg: EncryptConfiguration): Promise<DecoratedR
     cfg.encryptionInformation,
     cfg.policy,
     cfg.mimeType,
-    cfg.tdfSpecVersion ?? '4.3.0'
+    cfg.tdfSpecVersion
```
eugenioenko, line 404: The generateManifest would use 
```
schemaVersion: targetSpecVersion || tdfSpecVersion
```

And `tdfSpecVersion` is `4.3.0`. 

Removing the `4.3.0` is better,  does it fix a bug or its cleanup?

lib/tdf3/src/client/index.ts
```diff
@@ -381,6 +381,7 @@ export class Client {
       windowSize = DEFAULT_SEGMENT_SIZE,
       keyMiddleware = defaultKeyMiddleware,
       streamMiddleware = async (stream: DecoratedReadableStream) => stream,
+      tdfSpecVersion,
```
eugenioenko, line 384: Great catch! 🥇

eugenioenko (approved):
Changes look good. Left a small question for sanity check

lib/tdf3/src/tdf.ts
dmihalcik-virtru: Yeah, sorry there are a few small cleanups mixed in with the fix. The fix is just the change in lib/tdf3/src/client/index.ts

eugenioenko: Awesome! This is fantastic refactor :)

PR #545, opened 2025-04-15
feat(sdk): connect rpc client export

eugenioenko:
**Description**
Generates and exports connectrpc service definitions.

**PR Changes**
- Updated connect client generation
- Re-exported generated client rpc services from OpenTDF client
- Added connectrpc framework as dependency
- Re-exported connectrpc framework lib
- Updated platform.sh script to generate client
- added make `generate-platform` to run the generator
- Added example component in `web-app` 
- Updated readme.md to  with examples of how to use the Platform Client

**Consumer Usage Example**
This is an example of how the generated client can be used:
```javascript
const platform = new PlaformClient({
      authProvider: oidcClient,
      platformUrl: '/api',
});
try {
  await platform.v1.keyAccessServerRegistry.listKeyAccessServers({});
}
```
**Consumer Example with Auth Interceptor**
This is an example of how the generated client can be used without an AuthProvider and use a call Interceptor instead.
```typescript
const authInterceptor: Interceptor = (next) => async (req) => {
  req.header.set('Authorization', `Bearer ${accessToken}`);
  return await next(req); 
};

const client = new PlaformClient({
  interceptors: [authInterceptor], 
  platformUrl: '/api',
});
```

**Generating**
To generate the service definition run:

```bash
make generate-platform
````
This will clone the `platform` into the repository and execute the generators based on the exported protos in the platform services

**Details**
To allow proto buf genrators to work, the [platform](https://github.com/opentdf/platform.git) needs to be cloned into `lib/src/platform` diretory. The script `scripts/platform.sh` handles that.

`lib/buf.gen.yaml` is the configuration for the generator.  To execute the generator manually
```bash
cd lib
npx buf generate ../platform/service
```
This will generate `lib/src/platform` exports for each of the service.
Finally, `lib/src/platform.ts` re-exports those services and the connect rpc framework

**Follow up**
- update the sdk to use internally the generated rpc services


lib/buf.gen.yaml
```diff
@@ -0,0 +1,8 @@
+version: v2
+
+clean: true
+plugins:
+  - local: protoc-gen-es
+    out: src/platform
+    include_imports: true
+    opt: target=ts,import_extension=.js
```
eugenioenko, line 8: TODO: Add a comment here for `target=ts`

The consumers do need  `js+dts` so that the library can be used just with javascript. They are getting that, here is why:

Internally `ts` files are generated. When a build is made, those generated `ts` files are compiled in `dist/lib/src/platform` into  `js` + `dts`.
So for the library itself, we use generate and use the `ts` definitions. For consumers after the build, its `js + dts`

lib/package.json
```diff
@@ -52,6 +52,11 @@
       "types": "./dist/types/src/nanoindex.d.ts",
       "require": "./dist/cjs/src/nanoindex.js",
       "import": "./dist/web/src/nanoindex.js"
+    },
+    "./platform": {
```
eugenioenko, line 56: TODO: verify if the name `platform` is expected here. This means  the consumers will use the import as so
```
from "@opentdf/sdk/platform"
```

lib/src/platform.ts
eugenioenko, line 1: TODO: verify if this can be exported by service.

Something like 
```
export * as wellknown  from "./platform/wellknownconfiguration/wellknown_configuration_pb.js"
```

The tradeoff is that every service will need to be renamed, but it does help keeping this  more modular

scripts/platform.sh
```diff
@@ -7,9 +7,6 @@ rm -rf platform lib/src/platform
 git clone https://github.com/opentdf/platform.git
 
 # Generate Typescript code
-cmd=$(npx buf --version)
-if [ $? -ne 0 ]; then
-  npm install @bufbuild/protoc-gen-es @bufbuild/buf @connectrpc/protoc-gen-connect-es
```
eugenioenko, line 12: I've put this dependencies as `devDependencies` in the package.json. They do no need to be installed here

lib/package.json
```diff
@@ -72,13 +77,17 @@
     "watch": "(trap 'kill 0' SIGINT; npm run build && (npm run build:watch & npm run test -- --watch))"
   },
   "dependencies": {
+    "@connectrpc/connect": "^2.0.2",
```
eugenioenko, line 80: These have been added as dependencies and are re-exported in `lib/src/platform.ts`. This way we can ensure that the generated client services will work with the connectrpc web framework.

One alternative I've considered here is to declare them as `peerDependencies` and actually **not** re-export them. When someone consumes `@opentdf/sdk` they will get a warning that `@connectrcp` needs to be added to their `package.json`. 
I would have pushed for this option if the bundle size increased, but it has not. I think we are good with re-export

lib/src/platform.ts
```diff
@@ -0,0 +1,20 @@
...
+export * from './platform/policy/registeredresources/registered_resources_pb.js';
+export * from './platform/policy/resourcemapping/resource_mapping_pb.js';
+export * from './platform/policy/selectors_pb.js';
+export * from './platform/policy/subjectmapping/subject_mapping_pb.js';
+export * from './platform/policy/unsafe/unsafe_pb.js';
+export * from './platform/wellknownconfiguration/wellknown_configuration_pb.js';
+
+// export Connect RPC framework
+export * from '@connectrpc/connect-web';
+export * from '@connectrpc/connect';
```
dmihalcik-virtru, line 20: Shouldn't these be export as? It seems like this will result in a lot of overly simplified names.

dmihalcik-virtru (changes requested):
Overall, this seems like a good consolidation over the previous work.

Have you actually managed to make these work? If you want to test them, you'll probably want to add code to the integration test packages (in the cli and web-app folders).

lib/src/platform.ts
eugenioenko: > Shouldn't these be export as?
I've had that thought too, currently there is no name collision but there might be one eventually.  I'll update to use a "name" export. As a result,  the imports will look like 
```
import { wellknown.WellKnownService }from "@opentdf/sdk/platform"
```
```
export * wellknown from './platform/wellknownconfiguration/wellknown_configuration_pb.js';
```

eugenioenko:
> Have you actually managed to make these work? If you want to test them, you'll probably want to add code to the integration test packages (in the cli and web-app folders).

I've tested the wellknownconfig endpoint  with the local build to make sure it was exported and functional. (Thats how I got the example for the PR description)

As a follow up I do have to update the `web-app` to use the generated endpoints.  If needed, I can do it as part of this PR

@dmihalcik-virtru 
For the integration test packages, which integrations test do you want to see updated/added for this PR to be mergeable?

dmihalcik-virtru (approved):

web-app/src/components/ConnectRpcExample.tsx
```diff
@@ -0,0 +1,62 @@
...
+const transport = platformConnectWeb.createConnectTransport({
+  baseUrl: '/api',
+});
+
+const wellknownClient = platformConnect.createClient(wellknown.WellKnownService, transport);
+const kasClient = platformConnect.createClient(kas.AccessService, transport);
+const policyAttributesClient = platformConnect.createClient(
+  policyAttributes.AttributesService,
+  transport
+);
```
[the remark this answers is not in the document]
eugenioenko: @jrschumacher you are spot on here. So far this PR only generates and re-exports the connect services. The next step is to refactor the web-sdk to actually use the connect services internally.  Once that done, I feel like that's gonna be the moment when we can export the services in a centralized way. 

We could technically do it now, but it doesn't feel right to export the connect rpc services from a client that is actually not using them yet.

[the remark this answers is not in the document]
eugenioenko: @jrschumacher, updated the client export to use the improved client experience so now this is possible:
```
const client = new OpenTDF({
      authProvider: oidcClient,
      platformUrl: '/api',
});
try {
  await client.services.keyAccessServerRegistry.listKeyAccessServers({});
}
```

I need to add a few tests so that we can test at lest a few of the rpc methods to make sure the auth interceptors are working as expected. I've tested manually and they are :)

lib/src/opentdf.ts
```diff
@@ -350,6 +357,12 @@ export class OpenTDF {
         true,
         ['sign', 'verify']
       );
+
+    this.services = CreatePlatformClient({
```
dmihalcik-virtru, line 361: This isn't a great API since it:

- allocates greedily - most applications won't use most of these services 
- isn't tree shakeable - if an applicaiton doesn't use one or more services, or any service at all, esbuild or other tree shaking app bundlers will include its source code, as it is still statically reachable

Maybe hold off on this for now and let the examples use `import {CreatePlatformClient} from '@opentdf/sdk/platform'` for now, or something similar?

scripts/platform.sh
```diff
@@ -7,9 +7,6 @@ rm -rf platform lib/src/platform
...
-fi
-npx buf generate platform/service
+cd lib
+npx buf generate ../platform/service
```
dmihalcik-virtru, line 11: Can we invoke this in `lib/package.json` from prepack, so we don't have to commit the generated ts files to the repo?

web-app/src/components/ConnectRpcExample.tsx
```diff
@@ -0,0 +1,48 @@
...
+  };
+
+  const handlePolicy = async () => {
+    const response = await client.services.v1.attributes.listAttributes({});
+    setResult(response.attributes.map((s) => `${s}`).join(','));
+  };
+
+  return (
+    <>
+      <fieldset>
```
dmihalcik-virtru, line 33: Can you add test coverage to this in web-app/tests/tests?

dmihalcik-virtru (approved):

scripts/platform.sh
eugenioenko: So we do actually need to commit them to the repo because they are used by platform.ts internally for the types.  Otherwise platform.ts won't build

lib/src/opentdf.ts
eugenioenko: It's a good point that it won't be tree-shakeable. 

Without adding the services to OpenTDF it would look something like this:
```
import {createPlatformClient} from '@opentdf/sdk/platform'

const client = new OpenTDF({
      authProvider: oidcClient,
      platformUrl: '/api',
});

const platfrom = createPlatformClient({
      authProvider: oidcClient,
      platformUrl: '/api',
})
try {
  await platform.services.v1.keyAccessServerRegistry.listKeyAccessServers({});
}```

If somebody needs the platfrom and the openTDF they would need to create both. Not necessary a bad thing.


Before doing the update, let me check how much is actually being added in the bundle. I can check by verifying the size of the build of the web-app before and after

Comment (eugenioenko) [reply lib/src/opentdf.ts:361]: Alright, numbers are in. Yeah, its peaty much ends up adding the size of the `dist/platform` folder.

**before**
```dist/assets/index-DQKHxfIZ.js       348.63 kB │ gzip: 110.69 kB```

**after**
```dist/assets/index-CsYy-Uz_.js       611.58 kB │ gzip: 174.03 kB```

Its about 300Kb and 60Kb gzipped.

I think for now, for the consumer to instantiate a platform client besides the OpenTDF is a good compromise.
I will also update the `createPlatformClient` into a class, so that it can be used with the same pattern as the OpenTDF class, meaning:

```
const platfrom = new PlatformClient({
      authProvider: oidcClient,
      platformUrl: '/api',
})
```

web-app/src/components/ConnectRpcExample.tsx
eugenioenko: I've added  tests for the rpc calls in `/lib/tests/web/platform-rpc.test.ts` and they are run under the wtr.
This component is mostly used for debugging, it handles click to a button and makes an rpc call. The rpc calls are covered by the mentioned test before.

lib/src/opentdf.ts
eugenioenko: Refactored and updated the docs

dmihalcik-virtru (approved):

lib/src/opentdf.ts
dmihalcik-virtru: Without some kind of test it is hard to say, but I'm pretty sure most bundlers will only include for (most of) the new code if application code accesses it via the /platform import.

PR #555, opened 2025-04-22
fix(ci): use keycloack docker

eugenioenko:
chainguard keycloak is now private, use the regular dockerhub keycloak

### DIFF
--- a/.github/workflows/roundtrip/docker-compose.yaml
+++ b/.github/workflows/roundtrip/docker-compose.yaml
@@ -1,7 +1,6 @@
 services:
   keycloak:
-    # This is kc 24.0.1 with opentdf protocol mapper on board
-    image: cgr.dev/chainguard/keycloak@sha256:37895558d2e0e93ffff75da5900f9ae7e79ec6d1c390b18b2ecea6cee45ec26f
+    image: keycloak/keycloak:25.0
     restart: always
     command:
     - "start-dev"
--- a/.github/workflows/roundtrip/docker-compose.yaml
+++ b/.github/workflows/roundtrip/docker-compose.yaml
@@ -1,6 +1,6 @@
 services:
   keycloak:
-    image: keycloak/keycloak:25.0
+    image: keycloak/keycloak:24.0.5
     restart: always
     command:
     - "start-dev"


.github/workflows/roundtrip/docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
 services:
   keycloak:
     # This is kc 24.0.1 with opentdf protocol mapper on board
```
jakedoublev, line 3: Can you remove the stale comment on line 3 please?

.github/workflows/roundtrip/docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
...
-    image: cgr.dev/chainguard/keycloak@sha256:37895558d2e0e93ffff75da5900f9ae7e79ec6d1c390b18b2ecea6cee45ec26f
+    image: us-docker.pkg.dev/prj-infra-automation-ktbz/remote-proxy-dockerhub/keycloak/keycloak:24.0.5
```
elizabethhealy, line 4:
```suggestion
    image: keycloak/keycloak:25.0
```
dont want to reference our private proxy in opentdf

elizabethhealy (changes requested):

.github/workflows/roundtrip/docker-compose.yaml
eugenioenko: Thank you for the link to the right image :)

.github/workflows/roundtrip/docker-compose.yaml
eugenioenko: good catch!

jakedoublev (approved):

eugenioenko:
Hey @elizabethhealy , thank you for the catch. I've updated the docker image to the right one now. 
I used 24.0.5 image. With 25.0 the container wouldn't start and it times out.  Probably something to check on later.

The `cross-client-test v4.26` is failing because it still trying to load the changuard image (and I can't seem to locate the source). Is it building from previous release?

jakedoublev:
> Hey @elizabethhealy , thank you for the catch. I've updated the docker image to the right one now. I used 24.0.5 image. With 25.0 the container wouldn't start and it times out. Probably something to check on later.
> 
> The `cross-client-test v4.26` is failing because it still trying to load the changuard image (and I can't seem to locate the source). Is it building from previous release?

Yes. It's checking out the repo at that tag and finding the old image still being used by the repo as of that time.

elizabethhealy (approved):

PR #556, opened 2025-04-23
chore(ci): Replace chainguard keycloak image

dmihalcik-virtru:
These are no longer available for free tier users, let along unauthenticated ones


jakedoublev:
I believe this is a duplicate of #555

dmihalcik-virtru:
already merged in #555

PR #557, opened 2025-04-25
feat(sdk): get KASes list from platform when allowedKases list is not passed

eugenioenko:
## Motivation
To avoid malicious KASes, when a file is decrypted we want to give the platform administrators the ability to make sure that that KAS is within the platform's allowed KASes.

For this we will use the gRPC endpoint `/policy.kasregistry.KeyAccessServerRegistryService/ListKeyAccessServers` that will return the list of allowed KAS during decryption process and verify that the requested KAS is within that list.

## PR Changes
- adds `fetchKeyAccessServers()` function to get the allowed kas list
- adds `platformUrl` parameter to the various clients
- adds a safeguard to throw error when neither `platformUrl` nor `allowedKasList` is passed

## Minimizing Impact
The `platformUrl` is changing the clients API signature and is a breaking change. To minimize this breaking change here is the list of the scenarios and in which the consumers might get affected:

**SCENARIO 1**
> When `allowedKASEndpoints` is passed it 

Works out of the box because we don't need to reach out to `/ListKeyAccessServers`.  
The consumer is telling the sdk which are the KASes it allows.
```javascript
 const client = new OpenTDF({
      authProvider: oidcClient,
      defaultReadOptions: {
        allowedKASEndpoints: [c.kas],
      }
    });
```

**SCENARIO 2**
> When `ignoreAllowList` is passed 

Works out of the box because we don't need to reach out to `/ListKeyAccessServers`.
The consumer is telling the sdk to allow all KASes.
```javascript
 const client = new OpenTDF({
      authProvider: oidcClient,
      defaultReadOptions: {
        ignoreAllowList: true,
      }
    });
```

**SCENARIO 3**
> Neither `ignoreAllowList` or `allowedKASEnpoints` is passed. `platfromUrl` is passed

`platfromUrl` parameter is required so that sdk knows how to locate the platform `/ListKeyAccessServers`
The sdk will get the list of allowed KAS from the platform
```javascript
 const client = new OpenTDF({
      platformUrl: 'https://opentdf-platform.com', // <-- BREAKING CHANGE
      authProvider: oidcClient
    });
```

**SCENARIO 4**
> Neither `ignoreAllowList`, `allowedKASEnpoints` nor `platformUrl`  is passed

When calling `decrypt` it will fail with error message 
```
platformUrl is required when allowedKasEndpoints is empty
```

```javascript
 const client = new OpenTDF({
      authProvider: oidcClient
    });
```

**SCENARIO 5**
> When `kasEndpoint` is passed

This currently fails on decrypt because neither `allowList` or `platformUrl` is passed. There is a potential fix we can explore. It should be safe to assume that the `kasEndpoint` is a safe KAS. If the `allowedKasList` is empty, we can add the `kasEndpoint` to that list and the decrypt will work. 

```javascript
 const client = new OpenTDF({
      kasEndpoint: kasUrl,
      authProvider: oidcClient
    });
``` 

## Pending
- [ ] Finish the function `fetchKeyAccessServers` to actually load the KAS list, currently its returning the platformURL
- [ ] Update the tests to account for the new scenarios
- [ ] Add a warning in the console if KAS list is unsafe

### DIFF
--- a/lib/src/opentdf.ts
+++ b/lib/src/opentdf.ts
@@ -13,7 +13,12 @@ import {
   AssertionConfig,
   AssertionVerificationKeys,
 } from '../tdf3/src/assertions.js';
-import { type KasPublicKeyAlgorithm, OriginAllowList, isPublicKeyAlgorithm } from './access.js';
+import {
+  type KasPublicKeyAlgorithm,
+  OriginAllowList,
+  fetchKeyAccessServers,
+  isPublicKeyAlgorithm,
+} from './access.js';
 import { type Manifest } from '../tdf3/src/models/manifest.js';
 import { type Payload } from '../tdf3/src/models/payload.js';
 import {
@@ -87,6 +92,7 @@ export type CreateNanoTDFOptions = CreateOptions & {
 };
 
 export type CreateNanoTDFCollectionOptions = CreateNanoTDFOptions & {
+  platformUrl: string;
   // The maximum number of key iterations to use for a single DEK.
   maxKeyIterations?: number;
 };
@@ -136,6 +142,8 @@ export type CreateZTDFOptions = CreateOptions & {
 export type ReadOptions = {
   // ciphertext
   source: Source;
+  // Platform URL
+  platformUrl?: string;
   // list of KASes that may be contacted for a rewrap
   allowedKASEndpoints?: string[];
   // Optionally disable checking the allowlist
@@ -156,6 +164,9 @@ export type ReadOptions = {
 export type OpenTDFOptions = {
   // Policy service endpoint
   policyEndpoint?: string;
+
+  // Platform URL
+  platformUrl?: string;
 
   // Auth provider for connections to the policy service and KASes.
   authProvider: AuthProvider;
@@ -286,6 +297,7 @@ export type TDFReader = {
 // SDK for dealing with OpenTDF data and policy services.
 export class OpenTDF {
   // Configuration service and more is at this URL/connectRPC endpoint
+  readonly platformUrl: string;
   readonly policyEndpoint: string;
   readonly authProvider: AuthProvider;
   readonly dpopEnabled: boolean;
@@ -305,11 +317,15 @@ export class OpenTDF {
     disableDPoP,
     policyEndpoint,
     rewrapCacheOptions,
+    platformUrl,
   }: OpenTDFOptions) {
     this.authProvider = authProvider;
     this.defaultCreateOptions = defaultCreateOptions || {};
     this.defaultReadOptions = defaultReadOptions || {};
     this.dpopEnabled = !!disableDPoP;
+    if (platformUrl) {
+      this.platformUrl = platformUrl;
+    }
     this.policyEndpoint = policyEndpoint || '';
     this.rewrapCache = new RewrapCache(rewrapCacheOptions);
     this.tdf3Client = new TDF3Client({
@@ -333,8 +349,14 @@ export class OpenTDF {
   }
 
   async createNanoTDF(opts: CreateNanoTDFOptions): Promise<DecoratedStream> {
-    opts = { ...this.defaultCreateOptions, ...opts };
-    const collection = await this.createNanoTDFCollection(opts);
+    opts = {
+      ...this.defaultCreateOptions,
+      ...opts,
+    };
+    const collection = await this.createNanoTDFCollection({
+      ...opts,
+      platformUrl: this.platformUrl,
+    });
     try {
       return await collection.encrypt(opts.source);
     } finally {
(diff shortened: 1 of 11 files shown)


lib/src/opentdf.ts
```diff
@@ -286,6 +297,7 @@ export type TDFReader = {
 // SDK for dealing with OpenTDF data and policy services.
 export class OpenTDF {
   // Configuration service and more is at this URL/connectRPC endpoint
+  readonly platformUrl: string;
   readonly policyEndpoint: string;
```
dmihalcik-virtru, line 301: What is the difference between `platformUrl` and `policyEndpoint`?

lib/src/opentdf.ts
```diff
@@ -87,6 +92,7 @@ export type CreateNanoTDFOptions = CreateOptions & {
 };
 
 export type CreateNanoTDFCollectionOptions = CreateNanoTDFOptions & {
+  platformUrl: string;
```
dmihalcik-virtru, line 95: Can we just use the one from the client? It would seem odd to me that you would use multiple platforms with the same client

lib/tdf3/src/client/index.ts
```diff
@@ -531,8 +544,12 @@ export class Client {
       throw new ConfigurationError('AuthProvider missing');
     }
     const chunker = await makeChunkable(source);
-    if (!allowList) {
+    if (!allowList && this.allowedKases) {
       allowList = this.allowedKases;
+    } else if (this.platformUrl) {
+      allowList = await fetchKeyAccessServers(this.platformUrl);
```
dmihalcik-virtru, line 550: Suggestion: This and the other two awaits above (dpopKeys and makeChunkable) could be performed concurrently, e.g. by taking their promises and explicitly performing a await on the `all` of them

lib/src/opentdf.ts
eugenioenko: That's a great question.  At this point the difference is mostly semantics.

`policyEndpoint` is derived by subtracting `/kas` from the `kasEndpoint`
https://github.com/opentdf/web-sdk/blob/bdb5ba6402503c27016e26f38d6208924e6b6f74/lib/tdf3/src/client/index.ts#L309

Looking at the code base, `policyEndpoint` is guessed in a few places, but I can't seem to find a place where it is used.

If it's safe to assume that `policyEndpoint == platformUrl` I can update this PR to drop `platformUrl` and use `policyEndpoint` instead.  

With this current PR the main difference between the two is that `platfromUrl` is NEVER guessed, it needs to be set explicitly

lib/src/opentdf.ts
eugenioenko: We end up using the one from the client afaik here:
https://github.com/opentdf/web-sdk/blob/bdb5ba6402503c27016e26f38d6208924e6b6f74/lib/src/opentdf.ts#L358

This is added because otherwise typescript will not allow to pass in the field

lib/tdf3/src/client/index.ts
eugenioenko: This is a great suggestion for optimization. 🎉 

One thing I am not sure how to tie together is that fetchKeyAccessServer might also need the dpop headers. 
If the used authProvided requested the token with dpop the server will expect the dpop for the request.
So unwrap keys will need to await dpopKeys and also fetchKeyAccessServer too. 
Still doable, but it might end up `Promise.All([makeChunkable(), fetchKeyAccessServer])` without the `dpopKeys` because fetchkeyAccessServer might need it.

Is that the case? Do we need to add dpop for fetchKeyAccessServer too?

jrschumacher: This is a good optimization, but not for a vulnerability fix.

dmihalcik-virtru: Filed a follow-up, DSPX-1011

lib/src/opentdf.ts
dmihalcik-virtru: This is a side effect of using the deprecated nanotdf.Client class, since it has a required type with this name. We can remove that as part of standardizing on the 1.x API, and then remove this as a required parameter from the Collection helper object, or perhaps unify on something more abstract.

dmihalcik-virtru (approved):
Please focus on getting a working integration test for this feature via the multi-kas scenarios in opentdf/tests/xtest

lib/src/opentdf.ts
eugenioenko: @dmihalcik-virtru Followup sounds perfect on this.

`platformEndpoint` does not exactly fit for what it would be because it's not an endpoint. It's the URL where the platform is hosted. And we also have `policyEndpoint` which points to the same thing as `platformUrl` except policyEndpoint sometimes is guessed by removing `/kas` sufix from the `kasEndpoint`

I do agree, we need to cleanup  the public api for the  OpenTDF  classs

PR #565, opened 2025-04-29
feat(cli): Pass the platform url on decrypt, add the platform kas to the allowlist when fetching

elizabethhealy:
Ensure were adding the base platform kas to the allowlist when fetching from the registry
Pass the platform url correctly in the cli
Use the platform url from opts in nano decrypt (match behavior of tdf3)

### DIFF
--- a/lib/src/access.ts
+++ b/lib/src/access.ts
@@ -192,7 +192,9 @@ export async function fetchKeyAccessServers(
     }
     // if we get an error from the kas registry, throw an error
     if (!response.ok) {
-      throw new ServiceError(`unable to fetch kas list from [${req.url}], status: ${response.status}`);
+      throw new ServiceError(
+        `unable to fetch kas list from [${req.url}], status: ${response.status}`
+      );
     }
     const { keyAccessServers = [], pagination = {} } = await response.json();
     allServers.push(...keyAccessServers);
@@ -202,7 +204,7 @@ export async function fetchKeyAccessServers(
   const serverUrls = allServers.map((server) => server.uri);
   // add base platform kas
   serverUrls.push(`${platformUrl}/kas`);
-  
+
   return new OriginAllowList(serverUrls, false);
 }
 
--- a/lib/src/access.ts
+++ b/lib/src/access.ts
@@ -190,17 +190,19 @@ export async function fetchKeyAccessServers(
     } catch (e) {
       throw new NetworkError(`unable to fetch kas list from [${req.url}]`, e);
     }
-    if (response.ok) {
-      const { keyAccessServers = [], pagination = {} } = await response.json();
-      allServers.push(...keyAccessServers);
-      nextOffset = pagination.nextOffset || 0;
+    // if we get an error from the kas registry, throw an error
+    if (!response.ok) {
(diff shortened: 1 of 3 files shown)


elizabethhealy:
passing updated xtest with kao tamper tests enabled for latest commit https://github.com/opentdf/tests/actions/runs/14735661687

lib/src/opentdf.ts
```diff
@@ -529,7 +529,7 @@ class NanoTDFReader {
       dpopEnabled: this.outer.dpopEnabled,
       dpopKeys: this.outer.dpopKeys,
       kasEndpoint: this.opts.allowedKASEndpoints?.[0] || 'https://disallow.all.invalid',
-      platformUrl: this.outer.platformUrl,
+      platformUrl: this.opts.platformUrl || this.outer.platformUrl,
```
dmihalcik-virtru, line 532: If we make it required in the constructor, won't this always resolve to the opts' field

elizabethhealy: yes, it was mainly because this.opts.platform's type was string | undefined and platformUrl wanted just string. wasnt sure the best way around that in ts. what would you suggest?

lib/src/access.ts
```diff
@@ -190,17 +190,21 @@ export async function fetchKeyAccessServers(
     } catch (e) {
       throw new NetworkError(`unable to fetch kas list from [${req.url}]`, e);
     }
-    if (response.ok) {
-      const { keyAccessServers = [], pagination = {} } = await response.json();
-      allServers.push(...keyAccessServers);
-      nextOffset = pagination.nextOffset || 0;
+    // if we get an error from the kas registry, throw an error
+    if (!response.ok) {
+      throw new ServiceError(
```
eugenioenko, line 195: 🎉  nice catch!

lib/src/access.ts
```diff
@@ -190,17 +190,21 @@ export async function fetchKeyAccessServers(
...
+    allServers.push(...keyAccessServers);
+    nextOffset = pagination.nextOffset || 0;
   } while (nextOffset > 0);
 
-  if (!allServers.length) {
-    throw new ConfigurationError('There are no available KAS');
-  }
   const serverUrls = allServers.map((server) => server.uri);
+  // add base platform kas
+  serverUrls.push(`${platformUrl}/kas`);
```
eugenioenko, line 206: No need to update code here, its more of a few questions.


1) There is an edge case when the same url might end up twice in this list.  If the server has itself in the KAS db this will end up adding it twice. Probably not an issue and also probably fixable with a simple ```if (!serversUrls.includes(`${platfromUrl}/kas`) { ...todo}```

2) By adding the default platfromUrl/kas into the list it prevents the list from being empty. I am not sure if there is another mechanism this can be done but here is the scenario that can be used for system admin to disable decryption:

- server administrator removes all entries from kas list
- consumer tries to decrypt a file
- its rejected because there are no allowed kas

Probably not the right way to do it, and there might be a better mechanism for this. Still just wanted to point it out for consideration

eugenioenko (dismissed):
Code changes look good  🎉

lib/src/access.ts
elizabethhealy: good call, updated!

eugenioenko (approved):

PR #596, opened 2025-05-12
feat(sdk):  sdk to use connect rpc calls

eugenioenko:
**Motivation** 
We want to deprecate the Connect Gateway and fully migrate to using Connect RPC

**PR Changes**
Updates the sdk fetch calls to use Connect RPC client
- Rewrap
- ListKeyAccessServers
- PublicKey
- AttributeValuesByFqns

**Notes**
The `AttributeValuesByFqns` endpoint has been modified at some point, but the type definition for the response have not been updated in the sdk. Now that we are using the generated types, the discrepancy surfaced, and it was addressed in this pr.

**Bundle size**
The bundle size of the web-sdk has increased in about 70Kb gziped.
I've tried to create an   Internal PlatformClient that only uses the services we need but I only saw marginal bundle size decrease.  After using the rollup visualizer to analyze what is using the extra space in the bundle these are the results:
- `@bufbuild/protobuf/dist/esm` is using 250Kb. This is the connect rpc serialization library. It's required when we export the client
- `buf/validate/validate_pb.js` 70Kb. This is the validation layer exported when any service is created
- `platform` 110Kb are the protobuf services. When any of those are used, the `validate_pb.js` will be included as well  

So, even if we try to minimize the protobuf definitions and the exported services, the moment we export connect client we are adding at least 250+70Kbs into the bundle. 

**x tests**
The x-tests are not passing but that is expected. The older servers version do not support rpc?


lib/src/access.ts
```diff
@@ -29,59 +18,18 @@ export type RewrapResponse = {
...
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/json',
-    },
-    body: JSON.stringify(requestBody),
-  });
-
-  let response: Response;
-
+  const platformUrl = getHostFromEndpoint(url);
```
eugenioenko, line 24: To get the key for the unrwap request the `url` is coming from the tdf manifest. The url uses `/kas` as prefix `https://host.com/kas`.
The issue is that to create platform transport we require the url or the platform, not the kas.

So not really sure whats the right way to do this

lib/src/access.ts
```diff
@@ -29,59 +18,18 @@ export type RewrapResponse = {
...
-      headers: req.headers,
-      redirect: 'follow', // manual, *follow, error
-      referrerPolicy: 'no-referrer', // no-referrer, *no-referrer-when-downgrade, origin, origin-when-cross-origin, same-origin, strict-origin, strict-origin-when-cross-origin, unsafe-url
-      body: req.body as BodyInit,
+    return await platform.v1.access.rewrap({
+      signedRequestToken,
     });
   } catch (e) {
-    throw new NetworkError(`unable to fetch wrapped key from [${url}]`, e);
+    throw new NetworkError(`[${platformUrl}] [Rewrap] ${extractRpcErrorMessage(e)}`);
```
eugenioenko, line 31: I've simplified this different instance of error thrown. The `error.message` contains the [RPC error codes](https://grpc.io/docs/guides/status-codes/) and in some cases the HTTP status code as well

lib/src/platform.ts
```diff
@@ -118,8 +118,12 @@ export class PlatformClient {
  */
 function createAuthInterceptor(authProvider: AuthProvider): Interceptor {
   const authInterceptor: Interceptor = (next) => async (req) => {
+    const url = new URL(req.url);
+    const pathOnly = url.pathname;
+    // TODO RCP: discuss in the PR if this is how we want to do it
+    // Signs only the path of the url in the request
     const token = await authProvider.withCreds({
-      url: req.url,
+      url: pathOnly,
```
eugenioenko, line 126: Signing only the path of the request fixes Dpop issue. With http gateway, the DPOP verification expects the whole url, ex: `https://host.com/v2/rewap`. But for RPC request it only expecting the path.

lib/src/policy/api.ts
```diff
@@ -1,10 +1,13 @@
...
+import { NetworkError } from '../errors.js';
 import { AuthProvider } from '../auth/auth.js';
-import { rstrip } from '../utils.js';
-import { GetAttributeValuesByFqnsResponse, Value } from './attributes.js';
+import { extractRpcErrorMessage, rstrip } from '../utils.js';
+import { PlatformClient } from '../platform.js';
+import { Value } from './attributes.js';
+import { GetAttributeValuesByFqnsResponse } from '../platform/policy/attributes/attributes_pb.js';
 
+// TODO KAS: go over web-sdk and remove policyEndpoint that is only defined to be used here
```
eugenioenko, line 8: We are reaching a point where policyEndpoint can be replaced by platformUrl, the update will require fixing xtests afaik

lib/src/policy/api.ts
```diff
@@ -14,48 +17,49 @@ export async function attributeFQNsAsValues(
...
       console.log(`Missing value definition for [${fqn}]; is this a valid attribute?`);
       continue;
     }
-    if (av.attribute && !av.value.attribute) {
-      av.value.attribute = av.attribute;
+    if (value && av.attribute && !value?.attribute) {
+      value.attribute = av.attribute;
     }
-    values.push(av.value);
+    // TODO: update the mapping
```
eugenioenko, line 52: Where is this being used?  I haven't updated the mock server and all the tests are passing. Roundtrip test are passing as well.

The mapping here is not correct: value.attribute, value.grants and value.metadata does not have the right shape and probably needs to be fixed. Where is it used?

eugenioenko: fixed

lib/src/access.ts
```diff
@@ -301,3 +203,15 @@ export class OriginAllowList {
...
   }
 }
+
+/**
+ * Converts a KAS endpoint URL to a platform URL.
+ * If the KAS endpoint ends with '/kas', it returns the host url
+ * Otherwise, it returns the original KAS endpoint.
+ */
+function getHostFromEndpoint(endpoint: string): string {
+  // TODO RPC: find a better way to get the right url, otherwise just use the `origin` function
```
eugenioenko, line 213: When `Rewrap` is called the URL received from the TDF manifest has the KAS endpoint sufix: `http://platfrom.com/kas`. The issue is that RPC requires the URL of the platform and not the KAS to create the transport layer.

If the URL is `http://platform.com/kas` we can either remove `/kas` or take the origin. Both would return `http://platform.com`

But if the url is `http://platform.com/v2/kas` than the result will be different and technically we can have cases where either might be correct

Not really sure how to resolve this in a solid way.

strantalis: You should just have to strip off `/kas` and the hostname will be the KAS you need to call.

lib/src/platform.ts
strantalis: I think this is something we need to fix in the auth middleware.

lib/src/access.ts
ntrevino-virtru: Does that support KAS-only mode? (I only know about it from this mention)

https://docs.google.com/presentation/d/1XYVcviUbrHrsvlOGLZcJQQHN2IJTKyViE2_oR1mDpAU/edit?slide=id.g2d86a436258_0_6#slide=id.g2d86a436258_0_6

eugenioenko: Updated to strip of trailing `/` and `/kas`

lib/src/platform.ts
eugenioenko: I think so too. I've added a task in the backlog to at least add a good example for Auth Interceptor with Dpop

lib/src/policy/api.ts
eugenioenko: Maybe if we leave `policyEndpoint` in the `cli` no update to x-tests will be needed

lib/src/policy/api.ts
```diff
@@ -1,61 +1,48 @@
...
-      mode: 'cors',
-      credentials: 'same-origin',
-      headers: req.headers,
-      redirect: 'follow',
-      referrerPolicy: 'no-referrer',
-    });
-  } catch (e) {
-    throw new NetworkError(`network error [${req.method} ${req.url}]`, e);
-  }
+  platformUrl = getPlatformUrlFromKasEndpoint(platformUrl);
```
dmihalcik-virtru, line 14: Ooooh, I get it now. I was like, how can this work, since xtest is testing against the gRPC-only server we tagged in October. It works because we are using grpc, not json.

https://connectrpc.com/docs/protocol/#summary

So I think we don't need to use @connectrpc/web at all? I guess the only problem is `bytes` fields? Do we have any? I guess probably wrapped keys?

lib/src/utils.ts
```diff
@@ -139,3 +140,32 @@ export async function extractPemFromKeyString(keyString: string): Promise<string
...
+  let result = endpoint || '';
+  if (result.endsWith('/')) {
+    result = rstrip(result, '/');
+  }
+  if (result.endsWith('/kas')) {
+    result = result.slice(0, -4);
+  }
+  if (result.endsWith('/v2/rewrap')) {
+    result = result.slice(0, -10);
+  }
```
dmihalcik-virtru, line 169: Shouldn't these be flipped?
```suggestion
  if (result.endsWith('/v2/rewrap')) {
    result = result.slice(0, -10);
  }
  if (result.endsWith('/kas')) {
    result = result.slice(0, -4);
  }
```

lib/src/policy/api.ts
eugenioenko: Technically, we don't strictly have to use @connectrpc/web. However, what it provides goes beyond just the transport layer — it abstracts the encoding and decoding of protobuf wire format, manages transport headers, supports compression, handles error mapping, and, thanks to the generated clients, enables built-in request and response validation.

Without it, we would need to implement those parts ourselves and handle raw binary requests with manual encoding and decoding, which is certainly possible but more error-prone and time-consuming.

Additionally, since we provide generated clients for our consumers, it makes sense for us to consume the same client internally as well. Doing so ensures consistency and reduces confusion for consumers who expect the same client experience we use in-house.

lib/src/utils.ts
eugenioenko: Thats a good catch, within tests it hasn't been a problem but I can see how if the url is `/kas/v2/rewrap` we do want this flipped

lib/src/access.ts
```diff
@@ -29,59 +23,20 @@ export type RewrapResponse = {
...
-        throw new NetworkError(
-          `${req.method} ${req.url} => ${response.status} ${response.statusText}`
-        );
-    }
-  }
-
-  return response.json();
+  const platformUrl = getPlatformUrlFromKasEndpoint(url);
+
+  return await tryPromisesUntilFirstSuccess(
```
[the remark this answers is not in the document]
eugenioenko: That's a good point. I created the task for it that https://virtru.atlassian.net/browse/DSPX-1121

Our own PEPs most of the time  instantiate a new version of the client every time description needs to be done, in that case it might not be as effective as in an app where the client is instantiated once. Still really great to explore

Makefile
```diff
@@ -47,4 +47,7 @@ doc:
 	cd lib && npm ci && npm run doc
 
 generate-platform:
-	./scripts/platform.sh
\ No newline at end of file
+	./scripts/platform.sh
+
+doit:
+	cd lib && npm run build && npm pack && cd .. && cd web-app && npm run start
```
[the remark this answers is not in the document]
eugenioenko: Oops... this is leftover, I didn't mean to leave it here. Gonna remove

lib/src/access.ts
```diff
@@ -301,3 +163,24 @@ export class OriginAllowList {
...
+ * @param first First function returning a promise to try.
+ * @param second Second function returning a promise to try if the first fails.
+ */
+async function tryPromisesUntilFirstSuccess<T>(
+  first: () => Promise<T>,
+  second: () => Promise<T>
+): Promise<T> {
+  try {
+    return await first();
+  } catch {
```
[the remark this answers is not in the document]
eugenioenko: Oh, this is fantastic suggestion, thank you!

lib/src/access/access-fetch.ts
```diff
@@ -0,0 +1,202 @@
...
+    throw new NetworkError(`unable to fetch public key from [${pkUrlV2}]`, e);
+  }
+  if (!kasPubKeyResponseV2.ok) {
+    switch (kasPubKeyResponseV2.status) {
+      case 404:
+        throw new ConfigurationError(`404 for [${pkUrlV2}]`);
+      case 401:
+        throw new UnauthenticatedError(`401 for [${pkUrlV2}]`);
+      case 403:
+        throw new PermissionDeniedError(`403 for [${pkUrlV2}]`);
```
dmihalcik-virtru, line 181:
```suggestion
      case 404:
      case 401:
      case 403:
        throw new PermissionDeniedError(`${kasPubKeyResponseV2.status} for [${pkUrlV2}]`);
```

lib/tests/web/platform-rest.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+describe('Local Platform Legacy Rest tests', () => {
+  it(`it calls rest endpoint fetchWrappedKey`, async () => {
+    try {
+      const result = await fetchWrappedKey(
+        `${platformUrl}/v2/rewrap`,
+        { signedRequestToken: '' },
+        authProvider
+      );
+      expect(!!result).to.equal(true);
```
dmihalcik-virtru, line 31: What is this supposed to be testing? AFAICT setting the `roundtrip-test-response` header will short-circuit the mock server.

dmihalcik-virtru (approved):

lib/tests/web/platform-rest.test.ts
eugenioenko: > AFAICT setting the roundtrip-test-response header will short-circuit the mock server.

Correct, its testing that the rpc call is reaching the server. It's not testing what the server responds, that's covered by other integration tests

PR #610, opened 2025-05-28
feat: export connect rpc generated platform from package.json

eugenioenko:
**Motivation**
When creating RPC requests, we want to have the ability to define the `type` of the request. Connect RPC generated platform provides those types.
The issue is that typescript named exports do not export the types. Meaning this line  only ends up exporting the values from common. Not the types. 
```js
export * as common from './platform/common/common_pb.js';
```

The types are currently in the build already, just not accessible. This PR makes the types accessible

**PR changes**
Exports everything  from the generated client`/dist/lib/platform*` as `/platform/*` from the package.json. 
This allows for all the types to be accessible

Examples of how a type can be imported now 

```js
import { ActiveStateEnum } from '@opentdf/sdk/platform/common/common_pb.js';
```

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -39,9 +39,9 @@
       }
     },
     "node_modules/@bufbuild/protobuf": {
-      "version": "2.4.0",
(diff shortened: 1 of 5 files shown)


dmihalcik-virtru (approved):
Cleaner imports!

PR #623, opened 2025-06-10
feat(sdk): get kas public key from base key in wellknown config

eugenioenko:
**Motivation**
We want to ensures the Web SDK securely and consistently uses the "Base Key" from the WellKnown endpoint for cryptographic operations, provides a fallback when other keys are missing. 

--- 

**PR Changes**
- updates fetchPublicKey to read the Public Key from BaseKey in the Wellknown Configuration  and fallback to PublicKey
- Added integration tests to cover new functionality

---

**Use BaseKey flow**

**Scenario ZTDF **
 ztdf encrypt reaches out to WellKnown to get public key
- Base Key is RSA: encryption succeeds
- Base Key is ECC: encryption succeeds
- Base Key not set: encryption succeeds with PublicKey fallback

**Scenario Nano**
 nano encrypt reaches out to WellKnown to get public key.
- Base Key is RSA: encryption fails
- Base Key is ECC: encryption succeeds
- Base key not set: encryption succeeds with PublicKey fallback
---

**Example**
Here is an example of instantiating the client to use the base key feature
```ts
 const client = new OpenTDF({
      authProvider: oidcClient,
      platformUrl: `http://localhost:8080`
    });
```

### DIFF
--- a/lib/src/opentdf.ts
+++ b/lib/src/opentdf.ts
@@ -335,7 +335,7 @@ export class OpenTDF {
     this.tdf3Client = new TDF3Client({
       authProvider,
       dpopKeys,
-      kasEndpoint: 'https://disallow.all.invalid',
+      kasEndpoint: this.platformUrl || 'https://disallow.all.invalid',
       policyEndpoint,
     });
     this.dpopKeys =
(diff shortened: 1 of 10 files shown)


lib/src/opentdf.ts
```diff
@@ -335,7 +338,7 @@ export class OpenTDF {
     this.tdf3Client = new TDF3Client({
       authProvider,
       dpopKeys,
-      kasEndpoint: 'https://disallow.all.invalid',
+      kasEndpoint: this.platformUrl || 'https://disallow.all.invalid',
```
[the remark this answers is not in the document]
eugenioenko: Currently, even when platfromUrl is passed, the defaultKASEndpoint is still required
```
 const client = new OpenTDF({
      authProvider: oidcClient,
       platformUrl: `http://localhost:8080`,
      defaultCreateOptions: {
        defaultKASEndpoint: c.kas,
      }
    });
```

This change allows using the `platformUrl` as the fallback defaultKASEndpoint.  This way now OpenTDF can be instantiated with only AuthProvider and platfromUrl and be able to encrypt/decrypt. 
Its a developer experience enhancement

```
 const client = new OpenTDF({
      authProvider: oidcClient,
      platformUrl: `http://localhost:8080`,

    });
```

lib/tdf3/src/client/builders.ts
```diff
@@ -51,6 +51,7 @@ export type EncryptParams = {
   streamMiddleware?: EncryptStreamMiddleware;
   assertionConfigs?: AssertionConfig[];
   defaultKASEndpoint?: string;
+  useBasePublicKey?: boolean;
```
c-r33d, line 54: So we decided to not have an option that enables the base key. Instead what we plan to do is check if the base key is on the configuration, if we get no errors when retrieving the key we should use it. If not, fallback to using the default kases passed in.

We are mainly choosing to do it this way for the following reasons:
1. Allow admins to setup a base key and enable our sdks to prefer them
2. Not break existing testing infrastructure which pushes the release timeline.

eugenioenko: Updated the PR to not use the flag `useBasePublicKey`

PR #664, opened 2025-06-30
chore(ci): Fix deliver workflow permissions

el-virt:
`JamesIves/github-pages-deploy-action` action requires `contents: write` not `pages: write`


eugenioenko (approved):
👍

PR #694, opened 2025-07-22
feat(sdk): plain text policy support

sujankota:
Adds support for specifying policyType (EmbeddedEncrypted or EmbeddedText) when creating NanoTDFs via the SDK and CLI.

**Key Changes:**

- CLI: New --policyType option for NanoTDF encryption.
- SDK: policyType is now passed through all layers and used in the binary NanoTDF header.
- Default remains EmbeddedEncrypted if not specified.

**Impact:**
- Users can now embed policies as plain text (EmbeddedText) or encrypted (EmbeddedEncrypted).
- No breaking changes; backward compatible.

### DIFF
--- a/lib/package-lock.json
+++ b/lib/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@opentdf/sdk",
-  "version": "0.3.2",
+  "version": "0.4.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@opentdf/sdk",
-      "version": "0.3.2",
+      "version": "0.4.0",
       "license": "BSD-3-Clause-Clear",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


lib/src/opentdf.ts
```diff
@@ -759,14 +762,12 @@ class Collection {
...
-        this.encryptOptions = { ecdsaBinding: true };
-        break;
-      case 'gmac':
-        this.encryptOptions = { ecdsaBinding: false };
-        break;
-    }
+
+    // Initialize encryptOptions with policyType if provided
+    this.encryptOptions = {
+      ecdsaBinding: opts.bindingType === 'ecdsa',
```
dmihalcik-virtru, line 768: This will always have `encryptOptions.ecdsaBinding` defined, instead of letting it be undefined, which changes the defaulting behavior

PR #748, opened 2025-10-06
feat(core): initial obligations support in rewrap flow

jakedoublev:
WIP

### DIFF
--- a/lib/src/access.ts
+++ b/lib/src/access.ts
@@ -2,6 +2,7 @@ import { type AuthProvider } from './auth/auth.js';
 import { ServiceError } from './errors.js';
 import { RewrapResponse } from './platform/kas/kas_pb.js';
 import { getPlatformUrlFromKasEndpoint, validateSecureUrl } from './utils.js';
+import { base64 } from './encodings/index.js';
 
 import {
   fetchKasBasePubKey,
@@ -13,6 +14,15 @@ import { fetchWrappedKey as fetchWrappedKeysLegacy } from './access/access-fetch
 import { fetchKasPubKey as fetchKasPubKeyRpc } from './access/access-rpc.js';
 import { fetchKasPubKey as fetchKasPubKeyLegacy } from './access/access-fetch.js';
 
+/**
+ * Header value structure for 'X-Rewrap-Additional-Context`
+*/
+export type RewrapAdditionalContext = {
+  obligations: {
+    fqns: string[];
+  }
+}
+
 export type RewrapRequest = {
   signedRequestToken: string;
 };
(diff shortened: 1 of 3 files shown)


jakedoublev:
3/4 of the Sonarcloud issues are existing code I only touched, and the 4th asking for DRYness within the ZTDFReader and NanoTDFReader I would rather avoid because they satisfy the same interface but may differ in implementation as Nano is aligned with other ZTDF features like splits and streaming.

lib/src/opentdf.ts
```diff
@@ -634,11 +654,25 @@ class NanoTDFReader {
...
+  /**
+   * Returns obligations populated from the decrypt flow.
+   * If a decrypt has not occurred, attempts one to retrieve obligations.
+   */
+  async obligations(): Promise<string[]> {
+    if (this.requiredObligations) {
+      return this.requiredObligations;
+    }
+    await this.decrypt();
+    return this.requiredObligations ?? [];
```
c-r33d: Do we want to make required obligations an object by chance? In case we ever decide to send more information.

lib/src/platform.ts
```diff
@@ -99,6 +105,10 @@ export class PlatformClient {
       unsafe: createClient(UnsafeService, transport),
       wellknown: createClient(WellKnownService, transport),
     };
+
+    this.v2 = {
```
c-r33d: Shouldn't need this anymore since we should not be calling Authz for on-demand Obligations.

jakedoublev: I actually decided to intentionally leave this as our SDK should support calling auth service v2 even if we don't have a need for it within the obligations flow.

lib/src/access.ts
```diff
@@ -22,26 +32,52 @@ export type RewrapRequest = {
...
+  fulfillableObligationValueFQNs: string[]
+): string | undefined => {
+  if (!fulfillableObligationValueFQNs.length) return;
+
+  const context: RewrapAdditionalContext = {
+    obligations: {
+      fulfillableFQNs: fulfillableObligationValueFQNs.map((fqn) => fqn.toLowerCase()),
+    },
+  };
+  return base64.encode(JSON.stringify(context));
```
eugenioenko: `JSON.stringify` can throw.  Do we want to catch it and return undefined or re-throw a "nicer" error?

ntrevino-virtru (approved):
This definitely seems fine, and like it will work. From an API standpoint, I'm not sure I'm fond of mixing setting `this.requiredObligations` in with a call to `decrypt`. I think it could always be cleaned up later though.

ntrevino-virtru (approved):

PR #804, opened 2026-01-20
feat: Add jwt utils

elizabethhealy:
## Note
- lib/tdf3/src/crypto/jose/vendor/* is all code pulled from jose
- the additional jwt code we need to maintain is just the small shims around the jose code lib/tdf3/src/crypto/jose and the ~200 loc in jwt.ts for signing and verification
- all jwt claim logic and crit verification is in the jose helpers
- Upgrade path: since the helper files are vendored, a jose bump is usually seamless if those helpers don’t change. The update flow is just npm run vendor:jose-jwt-helpers + re-run tests. If jose changes
  helper file paths or signatures, we’ll need a small tweak to the vendoring script or wrappers, but this is contained and quick

## Summary of Changes

This pull request introduces a robust and secure set of utilities for handling JSON Web Tokens (JWTs) within the SDK. By integrating JWT signing and verification directly with the existing `CryptoService`, it centralizes cryptographic operations and enhances control over security primitives. The changes also involve vendoring necessary non-cryptographic components from the `jose` library to maintain functional parity while ensuring cryptographic operations are managed by the SDK's own services. This significantly improves the security posture and maintainability of JWT handling.

### Highlights

* **CryptoService Integration for JWTs**: New `signJwt` and `verifyJwt` functions leverage the SDK's `CryptoService` for all cryptographic operations (signing/verifying), moving away from `jose`'s internal `crypto.subtle` calls.
* **Vendoring of `jose` Helpers**: Non-cryptographic helper functions from the `jose` library (e.g., claims validation, crit handling, errors, utils) are now vendored via a build-time script to maintain `jose`'s semantics without relying on its WebCrypto usage.
* **Automated Vendoring Process**: A new `npm run vendor:jose-jwt-helpers` script has been added to `package.json` and integrated into the `build` process, ensuring that `jose` helper files are automatically copied and converted to TypeScript. The vendored directory is also added to `.prettierignore`.
* **Comprehensive JWT Unit Tests**: Extensive unit tests (`jwt.spec.ts`) have been introduced to thoroughly validate the behavior of the new `CryptoService`-based JWT signing and verification, covering various scenarios including symmetric (HS256) and asymmetric (RS256) algorithms, claim validations, and security checks.
* **Refined Key Extraction Logic**: Removed commented-out code related to `extractEcCurveFromPrivateKey` in `lib/tdf3/src/crypto/index.ts`, streamlining the crypto module.

### DIFF
--- a/lib/package-lock.json
+++ b/lib/package-lock.json
@@ -13,7 +13,7 @@
         "@connectrpc/connect-web": "^2.0.2",
         "buffer-crc32": "^1.0.0",
         "dpop": "^1.4.1",
-        "jose": "^6.0.8",
+        "jose": "6.0.8",
         "json-canonicalize": "^1.0.6",
         "uuid": "~11.1.0"
       },


eugenioenko (approved):
@elizabethhealy thank you for putting this together, this is a really elegant solution to our Jose problem. Really appreciate the vendoring script help as well, that will make upgrading easier in the future.

One thing I did notice is that `package.json` sets Jose with caret range`^`: 
```
 "jose": "^6.0.8"  // Allows 6.0.8 → 6.999.999
```

This might get problematic because it allows version bumping without modifying package.json, which could cause drift between the installed Jose version and vendored files.

As a follow-up ticket, we should pin the version:
```
 "jose": "6.0.8" 
```

Given the stacked PRs, let's handle this separately. Approving!

PR #907, opened 2026-04-02
chore: update root README with interceptor-first examples

eugenioenko:
## Summary
- Lead with the interceptor pattern (`authTokenInterceptor`, `authTokenDPoPInterceptor`, custom interceptors) as the recommended auth approach
- Move AuthProvider examples to a clearly marked "Deprecated since 0.14.0" section
- Add `authProviderInterceptor` bridge example for migration path
- Update PlatformClient section to use interceptors by default
- Remove the old "Authorization Middleware Options" subsections

Follow-up to #899.

## Test plan
- [ ] Verify README renders correctly on GitHub
- [ ] Verify code examples are syntactically correct

## Summary by CodeRabbit

* **Documentation**
  * Updated README to recommend an interceptor-first authentication pattern and show how to supply token interceptors for request/header handling
  * Added DPoP authentication examples and guidance for configuring associated keys via interceptors
  * Clarified client construction with a simplified example and explicit endpoint/platform configuration option
  * Documented continued support for the legacy AuthProvider approach (marked deprecated) and migration path

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,38 +8,23 @@ and some management tasks for ABAC.
 
 ## Usage
 
+### With Interceptors (Recommended)
+
+Use interceptors to provide authentication. The SDK does not manage tokens — you bring your own auth.
+
 ```typescript
-import { AuthProviders, OpenTDF } from '@opentdf/sdk';
-
-// Configuration Options
-const kasEndpoint = "http://localhost:65432/kas";
-
-// Authentication options (vary by middleware)
-const oidcOrigin = "http://localhost:65432/auth/realms/opentdf";
-const clientId = "applicationNameFromIdP";
-const refreshToken = "refreshTokenValueFromIdP";
-
-// AuthProviders are middlewares that add `Authorization` or other bearer tokens to requests.
-// These include The `refresh` provider can be handed a refresh and optional access token.
-const authProvider = await AuthProviders.refreshAuthProvider({
-  clientId,
-  exchange: 'refresh',
-  refreshToken,
-  oidcOrigin,
-});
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,142 +8,126 @@ and some management tasks for ABAC.
...
+```typescript
+import { authTokenInterceptor, OpenTDF } from '@opentdf/sdk';
 
 const client = new OpenTDF({
-  authProvider,
-  defaultCreateOptions: {
-    defaultKASEndpoint: kasEndpoint,
-  },
-  dpopKeys: authProvider.getSigningKey(),
+  interceptors: [authTokenInterceptor(() => myAuth.getAccessToken())],
```
marythought: instead of `myAuth.getAccessToken()` can we define this function above and then call `getAccessToken()`? something like: 
 
```
// see https://www.jwt.io for examples of encoded/decoded access tokens 
// implementation will vary depending on your auth provider
const getAccessToken = function() {
  return `my-access-token`;
}
```

or we could wait until https://github.com/opentdf/web-sdk/pull/906 is merged to use an actual function that we provide. I worry the placeholder and "myAuth" package will confuse people.

eugenioenko: > implementation will vary depending on your auth provider

Would `implementation will vary depending on you identity provider` be better?  (Just to avoid confusing with AuthProvider)

Regarding #906 906, the tricky portion is that client credentials  flow is design for private clients used in Node.js. This being the WebSDK private client would be confusing or imply that node is supported (which is not)

marythought (approved):
🚀

PR #912, opened 2026-04-09
fix(ci): allow main scope in PR title validation

marythought:
## Summary

- Adds `main` to the allowed scopes in the Conventional Commits Check
- Fixes `build-and-test / ccc` check failing on release-please PRs (e.g. #900)

Release-please uses `chore(main): release sdk ${version}` for PR titles, but `main` was not in the allowed scopes list.

## Test plan

- [x] Verify this PR's own `ccc` check passes
- [ ] Confirm #900 passes after this merges (or re-run its check)

### DIFF
--- a/.github/workflows/reusable_build-and-test.yaml
+++ b/.github/workflows/reusable_build-and-test.yaml
@@ -42,10 +42,12 @@ jobs:
           #   - ci: anything related to ci
           #   - tests: test only changes
           #   - docs: anything related solely to documentation
+          #   - main: release-please release PRs
           scopes: |
             ci
             cli
             docs
+            main
             sdk
             tests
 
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable_build-and-test.yaml
```diff
@@ -42,10 +42,12 @@ jobs:
           #   - ci: anything related to ci
           #   - tests: test only changes
           #   - docs: anything related solely to documentation
+          #   - main: release-please release PRs
           scopes: |
             ci
             cli
             docs
+            main
```
eugenioenko, line 50: *nit*: `chore:` would have been better for this, but release-please already uses `main:`

eugenioenko (approved):

.github/workflows/reusable_build-and-test.yaml
marythought: I believe Paul tried to get it to use `sdk` instead of `main` in a prior PR, but it is still using `main` for unexplainable reasons 🤷

PR #932, opened 2026-05-11
spec: scaffold for DSPX-3229

dmihalcik-virtru:
ml-kem kaos


dmihalcik-virtru:
Closed in favor of #933

PR #937, opened 2026-05-28
fix(sdk): allow duplicate KAOs on same KAS within a split (DSPX-3379)

dmihalcik-virtru:
## Summary
- `splitLookupTableFactory` no longer throws when a TDF's KAO array contains multiple entries with the same `sid` and KAS `url`. The inner value type is now `KeyAccessObject[]`, and duplicates are appended.
- `unwrapKey` registers one rewrap candidate per KAO in the list (disambiguating duplicate URLs with a `#<idx>` suffix), so `anyPool` tries each in turn — succeeding if any succeed, otherwise surfacing a normal permission failure.
- Unit tests updated to the new array shape; throw-on-duplicate test replaced with one that asserts duplicates are preserved; added coverage for "distinct kids on the same KAS for one split."

Fixes [DSPX-3379](https://virtru.atlassian.net/browse/DSPX-3379).

## Test plan
- [x] `make test` — lib mocha + WTR suites pass (334 passing), including the new `splitLookupTableFactory` cases. (web-app failure in local run was a missing Playwright Chromium binary, unrelated to this change.)
- [ ] Decrypt a TDF whose KAO array contains a duplicated `(sid, url)` pair end-to-end and confirm successful decrypt instead of the prior `InvalidFileError("Repetition found …")`.
- [ ] Decrypt one where all duplicates would be rejected by KAS and confirm a clean permission-denied error.

[DSPX-3379]: https://virtru.atlassian.net/browse/DSPX-3379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/tdf3/src/tdf.ts
+++ b/lib/tdf3/src/tdf.ts
@@ -706,7 +706,7 @@ export async function loadTDFStream(chunker: Chunker): Promise<InspectedTDFOverv
 export function splitLookupTableFactory(
   keyAccess: KeyAccessObject[],
   allowedKases: OriginAllowList
-): Record<string, Record<string, KeyAccessObject>> {
+): Record<string, Record<string, KeyAccessObject[]>> {
   const allowed = (k: KeyAccessObject) => allowedKases.allows(k.url);
   const splitIds = new Set(keyAccess.map(({ sid }) => sid ?? ''));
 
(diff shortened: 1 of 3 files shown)


dmihalcik-virtru:
Closed in favor of #967

PR #952, opened 2026-06-22
fix(tdf3): remove TODO from customer-facing error message

davidaronhopper-dev:
## Summary
Fixes DSPX-3454 - removes TODO from customer-facing error message in Secure Viewer.

A customer reported seeing: `"TODO: Fallback to no split ids. Repetition found for..."` in error output when decryption failed.

## Changes
- Replaced customer-facing error with user-friendly message: `"Unable to decrypt: Multiple keys detected for Key Access Server [URL]. Please contact your administrator."`
- Kept developer TODO as code comment for future implementation of split ID fallback logic

## Impact
- Customers no longer see internal TODO messages in error output
- Developer context preserved in code comments for future work

## Related
- Follow-up ticket created: DSPX-3666 (audit codebase-wide for other customer-facing TODOs)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation error messaging for split key-to-URL configuration issues. If multiple keys map to the same URL within a split, the error now clearly reports “Multiple keys detected” and includes guidance to contact your administrator.  
* **Tests**
  * Updated the unit test expectations to match the revised error message.

### DIFF
--- a/lib/tdf3/src/tdf.ts
+++ b/lib/tdf3/src/tdf.ts
@@ -731,7 +731,7 @@ export function splitLookupTableFactory(
       // A proper fix would handle this by deduplicating or selecting the appropriate key.
       throw new InvalidFileError(
         `Unable to decrypt: Multiple keys detected for Key Access Server [${kao.url}]. ` +
-        `Please contact your administrator.`
+          `Please contact your administrator.`
       );
     }
     if (allowed(kao)) {
--- a/lib/tdf3/src/tdf.ts
+++ b/lib/tdf3/src/tdf.ts
@@ -726,8 +726,12 @@ export function splitLookupTableFactory(
   for (const kao of keyAccess) {
     const disjunction = splitPotentials[kao.sid ?? ''];
     if (kao.url in disjunction) {
+      // TODO(DSPX-3454): Implement fallback to no split ids when repetition is detected.
+      // Repetition occurs when multiple KAS keys share the same URL within a split.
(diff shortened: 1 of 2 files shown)


lib/tdf3/src/tdf.ts
```diff
@@ -726,8 +726,12 @@ export function splitLookupTableFactory(
   for (const kao of keyAccess) {
     const disjunction = splitPotentials[kao.sid ?? ''];
     if (kao.url in disjunction) {
+      // TODO(DSPX-3454): Implement fallback to no split ids when repetition is detected.
```
ntrevino-virtru: If each KAO has the KID I think it should be fine. It seems like the TODO should be to update this function  with that knowledge? And we probably need some xtest test added to validate how it works across SDKs currently.

davidaronhopper-dev: From what JP said in slack, this ticket didn't really need a "fix" it just needed to have the TODO removed from anything client facing, I can update the PR if that isn't all that should be done, I also have another PR to check the whole codebase for other TODOs in a similar vein

ntrevino-virtru: Yeah, I mean I don't want to misdirect with the TODO. I'm pretty sure using no split ids is not correct.

davidaronhopper-dev: updated per your comment

dmihalcik-virtru (approved):

PR #1000, opened 2026-08-20
feat(main): surface wrap and rewrap algorithm choice in web-app demo (DSPX-3229)

dmihalcik-virtru:
## What

Makes the ML-KEM key-wrap work from DSPX-3229 legible in the browser sample app, so it can be
demonstrated (and recorded) end to end. No `lib/` changes — this is all `web-app/`.

Three commits:

1. **`feat(main)`: surface wrap and rewrap algorithm choice.** The Manifest Inspector only
   populated on decrypt, so choosing ML-KEM and hitting Encrypt left nothing on screen to show it
   worked. `createZTDF` returns a `DecoratedStream` whose `manifest` promise is already resolved,
   so the encrypt-side inspector is free — no re-parse of the output. Also adds a **rewrap key
   algorithm** selector: `ReadOptions.wrappingKeyAlgorithm` picks the client's ephemeral keypair
   for the rewrap exchange, and left unset it falls through to RSA-2048. Without a control, half
   the quantum-safe story — the DEK's journey *back* from the KAS — stays classical and invisible.

   Also adds `.github/workflows/roundtrip/.gitignore`. The root one anchors `/platform` and
   `/*.pem` to the repo root, so local backend bring-up artifacts (the platform symlink, generated
   KAS private keys, the Keycloak CLI zip) would otherwise be snapshotted into the working copy.

2. **`style(main)`: restructure the layout for narrow displays.** The page read as one wide row of
   unrelated fieldsets, and the inspector was a six-column table that wrapped into itself below
   about a laptop width. Now: vertical is flow, horizontal is detail.

   ```
   source + sink  ->  encrypt | decrypt  ->  output + manifest
   ```

   Each action carries the algorithm that affects it, so the wrap selector sits with Encrypt and
   the rewrap selector with Decrypt. Encrypt and decrypt pair up on one row when there's room and
   stack when there isn't. Inspector table becomes label/value rows per KAO.

3. **`feat(main)`: tag download names with the wrap and rewrap mechanisms.**

   | Step | Before | After |
   |---|---|---|
   | Encrypt (`ec:secp256r1`) | `README.md.tdf` | `README.md-ecsecp256r1.tdf` |
   | Encrypt (`mlkem:768`) | `README.md.tdf` | `README.md-mlkem768.tdf` |
   | Decrypt (rewrap `mlkem:1024`) | `README.decrypted.md` | `README-mlkem768-rwk-p=mlkem1024.decrypted.md` |

   A folder of demo output now says which encapsulation produced which file, on both legs.

## Why

`mlkem:768` / `mlkem:1024` are available today but not yet the default. The demo needs to make two
things concrete that were previously invisible: that the KAO wrap really is post-quantum (visible
`type: mlkem-wrapped`, `kid`, and the 1158 / 1638-byte `wrappedKey`), and that the rewrap leg is a
*separate* choice most people miss.

## How to test

Bring up the roundtrip harness backend (see `.github/workflows/roundtrip/`), then:

```sh
cd web-app/tests && npx playwright test --reporter=line --workers=1
```

18 passed locally (3 browsers x 6 tests). `--workers=1` sidesteps a **pre-existing** race in
`Remote Source Streaming`: `static-server.js` binds port 8086 with no EADDRINUSE handling and all
three browser projects share `tests/README.md.tdf`. Confirmed pre-existing by stashing these
changes and reproducing 3/3 failures on the parent commit.

Test changes:
- The ML-KEM loop now asserts `#kao-kid-0` / `#kao-type-0` / `#kao-wrapped-bytes-0` immediately
  after Encrypt (before any decrypt), and drives `#rewrapAlgorithm` so the PQ rewrap leg is
  covered.
- New `download names record both key wrap legs` re-uploads the file *under the name encrypt
  suggested*. The other tests hand back Playwright's temp UUID, so the filename parser's happy
  path was never actually exercised — only its fallback branch.

Also `npm run lint`, `npm test` (Vitest), and `tsc --noEmit` clean in `web-app`.

## Risk

Low, and confined to the sample app. No `lib/` or `cli/` changes, no change to the TDF container,
policy model, or KAS protocol shape. The one behavioural change users will notice is the download
filenames.

Note the decrypt name keeps the existing `.decrypted` infix rather than dropping it, since it's
what distinguishes output from input in the same folder.

## Screenshots

Wide:
image

Narrow: 

image

## Summary by CodeRabbit

* **New Features**
  * Added separate encryption and rewrap algorithm controls.
  * Downloaded files now include applicable algorithm details in their filenames.
  * Added a Manifest Inspector displaying key access metadata after encryption and decryption.
  * Improved responsive layouts for workflows, cards, status elements, and action sections.

* **Bug Fixes**
  * Switching input sources now clears previous downloads, metadata, and warnings.
  * Improved filename handling and warnings when selected and actual algorithms differ.

* **Tests**
  * Expanded coverage for filenames, metadata inspection, rewrap scenarios, and input-source changes.

### DIFF
--- a/.github/workflows/roundtrip/.gitignore
+++ b/.github/workflows/roundtrip/.gitignore
@@ -5,12 +5,21 @@
 # Platform checkout (CI clones it here; locally it is usually a symlink)
 /platform
 
-# KAS keys and CA material from ./init-temp-keys.sh
+# KAS keys and CA material from ./init-temp-keys.sh. The CA serial lands in
+# keys/ alongside the cert it numbers, so /keys/ already covers it.
 /*.pem
-/*.srl
 /ecparams.tmp
 /keys/
 
 # Keycloak admin CLI downloaded by ./config-demo-idp.sh
 /kc.zip
 /keycloak-*/
+
+# ./wait-and-test.sh installs the packed CLI tarball here; only package.json is
+# tracked, so the lock file it generates is throwaway.
+/package-lock.json
+
+# Round-trip fixtures from ./encrypt-decrypt.sh. It only removes these on the
+# success path, and it is `set -e`, so a failed run strands them.
+/sample*.txt
+/sample*.tdf
--- a/.github/workflows/roundtrip/.gitignore
+++ b/.github/workflows/roundtrip/.gitignore
@@ -1,0 +1,16 @@
+# Local backend bring-up artifacts. In CI these live on a throwaway runner, but
+# when running the harness locally they land in this directory. See the
+# roundtrip README steps in web-app/tests/README.md.
+
+# Platform checkout (CI clones it here; locally it is usually a symlink)
+/platform
+
+# KAS keys and CA material from ./init-temp-keys.sh
+/*.pem
+/*.srl
+/ecparams.tmp
+/keys/
+
+# Keycloak admin CLI downloaded by ./config-demo-idp.sh
+/kc.zip
+/keycloak-*/
--- a/web-app/src/App.tsx
+++ b/web-app/src/App.tsx
@@ -3,7 +3,14 @@ import { useState, useEffect, type ChangeEvent } from 'react';
 import streamsaver from 'streamsaver';
 import { showSaveFilePicker } from 'native-file-system-adapter';
 import './App.css';
-import { type Chunker, type KasPublicKeyAlgorithm, type Source, OpenTDF } from '@opentdf/sdk';
+import {
+  type Chunker,
+  type DecoratedStream,
+  type KasPublicKeyAlgorithm,
+  type Manifest,
+  type Source,
+  OpenTDF,
+} from '@opentdf/sdk';
 import { type SessionInformation, OidcClient } from './session.js';
 import { config } from './config.js';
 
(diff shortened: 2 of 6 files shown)


eugenioenko (dismissed):
🎉

.github/workflows/roundtrip/.gitignore
pflynn-virtru, line 1: can we use the project root `.gitignore` ?

dmihalcik-virtru: I'd rather not clutter it. This is to ignore all the junk that the CI jobs create, which includes checking out a copy of o/platform here and running its scripts.

web-app/src/App.tsx
```diff
@@ -587,182 +628,220 @@ function App() {
...
+                <div id="downloadState" className="status">
+                  {downloadState}
+                </div>
+              )}
+              {algorithmWarning && (
+                <div id="algorithmWarning" className="status warning" role="alert">
+                  ⚠️ {algorithmWarning}
+                </div>
+              )}
+              {kaoMetadata?.length ? (
```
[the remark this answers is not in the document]
dmihalcik-virtru: Fixed in 4b26524.

Confirmed real: `setFileHandler` / `setRandomHandler` / `setUrlHandler` only ever called `setInputSource`, and the `#clearFile` button that did the full reset is rendered only when `hasFileInput` is true — so url and random-byte sources had no reset path at all.

All four call sites now go through one `selectInputSource` helper that clears `downloadState`, `kaoMetadata` and `algorithmWarning` alongside `inputSource`. Deliberately not a `useEffect` on `inputSource`: the handlers allocate a fresh object per keystroke, so an effect would also fire on mount and on identity churn.

Pinned with a new Playwright test, `changing the source retires the previous output`, which encrypts from a random-bytes source and then changes the byte count. Verified it reproduces the bug — reverting the helper fails it with `locator resolved to <div class="status" id="downloadState">✍️ Encrypt Complete</div>` — and passes with the fix. Full roundtrip suite: 21 passed (7 tests × 3 browsers) against a local platform.

pflynn-virtru (approved):

PR #1002, opened 2026-08-24
feat(sdk): resolve attribute key splits via GetKeyMappingsByFqns during encrypt

alkalescent:
## What / Why

The policy RPC `GetAttributeValuesByFqns` is deprecated in favor of narrower read APIs. For the client-side encrypt (autoconfigure / key-split) path the replacement is `GetKeyMappingsByFqns`, which returns the server-resolved effective KAS keys and attribute rule per value FQN (value > definition > namespace). This migrates the Web SDK encrypt path onto it, matching the platform Go SDK which already resolves splits this way.

## Changes

- **Proto bindings** (`fix(sdk)` commit): regenerate the vendored `lib/src/platform` bindings against `protocol/go/v0.41.0` so the attributes service exposes `getKeyMappingsByFqns` (added upstream in v0.37.0; the previous pin was v0.33.1). Mirrors the automated proto-update workflow.
- **Feature** (`feat(sdk)` commit):
  - `lib/src/policy/api.ts`: new `attributeFQNsAsKeyMappings` calls `getKeyMappingsByFqns` and adapts each mapping into a `Value` carrying `kasKeys` (consumed unchanged by `effectiveKasKeys` / `plan`). Values with no mapped keys (legacy KAS grants) fall back to `attributeFQNsAsValues` for just those FQNs; an `Unimplemented` response falls back for all FQNs.
  - `lib/tdf3/src/client/index.ts`: the `Client.encrypt()` autoconfigure branch now calls `attributeFQNsAsKeyMappings`.
  - Tests: `GetKeyMappingsByFqns` handler added to the mock server; `attributeFQNsAsKeyMappings` coverage in `platform-rpc.test.ts`.

## Fallback Order

Mapped keys from `GetKeyMappingsByFqns`, then legacy grants via `GetAttributeValuesByFqns` for empty-key FQNs, then the existing default-KAS behavior. An empty key set never bypasses the grant fallback.

## How to Test

`cd lib && npm run build && npm run lint && npm test`. Verified locally: build, eslint, mocha, and the full web-test-runner suite (253 passed, coverage 74.45%).

## Risk

Touches the encrypt key-split path. No change to produced TDFs for existing policy configurations: mapped-key values resolve to the same KAS keys server-side, and legacy-grant values take the unchanged `GetAttributeValuesByFqns` path.

## Summary by CodeRabbit

- **New Features**
  - Added dynamic value mapping with create, update, list, retrieve, and delete operations.
  - Added key-mapping and entitlement lookup APIs for automatic attribute resolution.
  - Added support for ML-KEM-768 and ML-KEM-1024 algorithms.
  - Added subject mappings when creating attribute values.
  - Added combined provider name and manager identification.

- **Improvements**
  - Attribute lookups now fall back gracefully when key-mapping services are unavailable.
  - Added key-status sorting and expanded key update capabilities.
  - Deprecated the legacy attribute-value lookup in favor of targeted lookup APIs.


alkalescent:
Addressed both review findings (9ba6cc3):

- **Update the autoconfigure test mock (P1):** `client.spec.ts` now stubs `GetKeyMappingsByFqns` (the RPC the encrypt path calls first) and asserts the mapped-key path. This also surfaced two `kas-key-splits.spec.ts` cases that stub fetch by call index; their grant/cached-key fixtures now return an empty key-mappings response so resolution falls back to `GetAttributeValuesByFqns`, with the fetch indices shifted for the extra RPC.
- **Fall back for missing map entries (P2):** `attributeFQNsAsKeyMappings` now iterates the requested FQNs instead of the response map, so a value the server omits falls back to `GetAttributeValuesByFqns` rather than being silently dropped — matching the platform Go SDK granter.

Local verification: `npm run build`, `npm run lint`, mocha 366/366, full web-test-runner suite all passing (coverage 74.51%).

lib/src/platform/policy/objects_pb.ts
```diff
@@ -1696,6 +1818,16 @@ export enum Algorithm {
...
+
+  /**
+   * @generated from enum value: ALGORITHM_MLKEM_768 = 20;
+   */
+  MLKEM_768 = 20,
+
+  /**
+   * @generated from enum value: ALGORITHM_MLKEM_1024 = 21;
+   */
+  MLKEM_1024 = 21,
```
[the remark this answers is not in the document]
alkalescent: fixed

lib/src/policy/api.ts
```diff
@@ -3,9 +3,78 @@ import { type AuthConfig, resolveInterceptors } from '../auth/interceptors.js';
...
+): Promise<Value[]> {
+  platformUrl = getPlatformUrlFromKasEndpoint(platformUrl);
+  const platform = new PlatformClient({ interceptors: resolveInterceptors(auth), platformUrl });
+
+  let response: GetKeyMappingsByFqnsResponse;
+  try {
+    response = await platform.v1.attributes.getKeyMappingsByFqns({ fqns });
+  } catch (e) {
+    if (e instanceof ConnectError && e.code === Code.Unimplemented) {
+      return attributeFQNsAsValues(platformUrl, auth, ...fqns);
```
elizabethhealy: attributeFQNsAsValues creates another platform client, should we modify to take the existing one instead?

lib/src/policy/api.ts
```diff
@@ -3,9 +3,78 @@ import { type AuthConfig, resolveInterceptors } from '../auth/interceptors.js';
...
+    throw new NetworkError(`[${platformUrl}] [GetKeyMappingsByFqns] ${extractRpcErrorMessage(e)}`);
+  }
+
+  const values: Value[] = [];
+  const legacyGrantFqns: string[] = [];
+  // Iterate the requested FQNs (not the response map) so a value the server omits
+  // still falls back rather than being silently dropped. The response is keyed by
+  // normalized (lower-cased) FQN.
+  for (const fqn of fqns) {
+    const mapping = response.fqnKeyMappings[fqn] ?? response.fqnKeyMappings[fqn.toLowerCase()];
```
elizabethhealy: are the fqnkeymappings guaranteed to be lower-cased /normalized from the server?

elizabethhealy (dismissed):

lib/src/policy/api.ts
alkalescent: Yes, via `resolveValueFqns` performs `strings.ToLower(fqn)` and keys the response map using that, so `fqnKeyMappings keys` are always lowercased. Simplified the lookup to `fqn.toLowerCase()` and updated the mock server to match."

lib/src/policy/api.ts
alkalescent: Refactored so `attributeFQNsAsKeyMappings` reuses its single `PlatformClient`.

elizabethhealy (approved):

PR #1003, opened 2026-08-25
docs(sdk): correct stale API comments

eugenioenko:
## Description

Correct stale TypeDoc comments that no longer match the SDK implementation. The updates cover ignored options, authentication requirements, KAS allowlist behavior, stream return types, lifecycle methods, parameter names, failure behavior, and an invalid usage example.

## Validation

- Prettier
- TypeDoc generation
- `git diff --check`

TypeDoc completes with zero errors. Existing unrelated warnings remain.

## Summary by CodeRabbit

* **Documentation**
  * Clarified authentication, client key setup, and signed request guidance.
  * Documented public key fetch error behavior when all attempts fail.
  * Updated guidance on unused signer and concurrency settings.
  * Clarified platform URL, allowlist, constructor, encryption, and decryption requirements.
  * Updated examples and resource cleanup behavior for readers and clients.


lib/src/access.ts
```diff
@@ -34,10 +34,9 @@ export type RewrapRequest = {
 /**
  * Get a rewrapped access key to the document, if possible
  * @param url Key access server rewrap endpoint
- * @param requestBody a signed request with an encrypted document key
- * @param authProvider Authorization middleware
+ * @param signedRequestToken Signed request token containing the encrypted document key.
+ * @param auth Connect interceptors or a legacy auth provider.
```
eugenioenko, line 38: The previous tags named `requestBody` and `authProvider`, which are not parameters, and included a nonexistent `clientVersion` tag. TypeDoc could not associate them with this signature.

lib/src/access.ts
```diff
@@ -209,7 +208,7 @@ export async function fetchECKasPubKey(kasEndpoint: string): Promise<KasPublicKe
  * This function will first try to fetch the base public key,
  * then it will try to fetch the public key using the RPC method,
  * and finally it will try to fetch the public key using the legacy method.
- * If all attempts fail, it will return the error from RPC Public Key fetch.
+ * If all attempts fail, throws the legacy public-key fetch error.
```
eugenioenko, line 211: The fallback helper throws the second, legacy-fetch error when both attempts fail. The previous comment incorrectly said the RPC error was returned.

lib/src/auth/providers.ts
```diff
@@ -26,9 +26,7 @@ import * as defaultCryptoService from '../../tdf3/src/crypto/index.js';
  * This just expects a clientId and clientSecret to be provided in the clientConfig, and will use that
  * to grant tokens via the OIDC clientCredentials flow.
  *
- * The client's public key must be set in all OIDC token requests in order to recieve a token with valid
- * Virtru claims. The public key may be passed to this provider's constructor, or supplied post-construction by calling
- * {@link updateClientPublicKey} which will force an explicit token refresh
+ * Bind the client public key after construction with `updateClientPublicKey()`.
```
eugenioenko, line 29: This factory does not accept or forward a public key. The previous comment described an unsupported constructor input.

lib/src/auth/providers.ts
```diff
@@ -59,9 +57,7 @@ export const clientSecretAuthProvider = async (
  * The client can supply this provider with a JWT issued by that trusted 3rd-party IdP, and that JWT will be exchanged
  * for a tokenset with TDF claims.
  *
- * The client's public key must be set in all OIDC token requests in order to recieve a token with valid
- * Virtru claims. The public key may be passed to this provider's constructor, or supplied post-construction by calling
- * {@link updateClientPublicKey}, which will force an explicit token refresh.
+ * Bind the client public key after construction with `updateClientPublicKey()`.
```
eugenioenko, line 60: This factory does not accept or forward a public key. The previous comment described an unsupported constructor input.

lib/src/auth/providers.ts
```diff
@@ -89,9 +85,7 @@ export const externalAuthProvider = async (
  * IdP using the OIDC flow fo their choice, and can provide a Refresh Token which will be exchanged (along with the client pubkey)
  * for a new tokenset containing valid TDF claims.
  *
- * The client's public key must be set in all OIDC token requests in order to recieve a token with valid
- * Virtru claims. The public key may be passed to this provider's constructor, or supplied post-construction by calling
- * {@link updateClientPublicKey} which will force an explicit token refresh
+ * Bind the client public key after construction with `updateClientPublicKey()`.
```
eugenioenko, line 88: This factory does not accept or forward a public key. The previous comment described an unsupported constructor input.

lib/src/opentdf.ts
```diff
@@ -78,7 +78,7 @@ export type CreateOptions = {
   /** The KAS to use for creation, if none is specified by the attribute service. */
   defaultKASEndpoint?: string;
 
-  /** Private (or shared) keys for signing assertions and bindings. */
+  /** Currently unused; supplying signer keys has no effect. */
```
eugenioenko, line 81: `createTDF()` and `createZTDF()` never read or forward `signers`. The previous comment implied that this option affected signing.

lib/src/opentdf.ts
```diff
@@ -152,7 +152,7 @@ export type ReadOptions = {
   /** Optionally disable assertion verification. */
   noVerify?: boolean;
 
-  /** If set, prevents more than this number of concurrent requests to the KAS. */
+  /** Currently ignored; KAS request concurrency is fixed at 1. */
```
eugenioenko, line 155: The reader ignores this option and passes `concurrencyLimit: 1`. The previous comment described behavior the option does not control.

lib/src/opentdf.ts
```diff
@@ -232,7 +232,7 @@ export type TDFReader = {
    */
   decrypt: () => Promise<DecoratedStream>;
   /**
-   * Mark this reader as closed and release any resources, such as open files.
+   * Marks the reader as closed. Does not release resources held by the source.
```
eugenioenko: The underlying reader close method is empty. The previous comment promised resource cleanup that is not implemented.

lib/src/opentdf.ts
```diff
@@ -256,8 +256,8 @@ export type TDFReader = {
  * The main OpenTDF class that provides methods for creating and reading TDF files.
  * It can be used to create new TDF files and read existing ones.
  * This class is the entry point for using the OpenTDF SDK.
- * It requires an authentication provider to be passed in the constructor.
- * It also requires a platform URL to be set, which is used to fetch key access servers and policies.
+ * Configure authentication with interceptors or the deprecated `authProvider`.
+ * Reads require a platform URL unless a KAS allowlist is supplied or allowlist checks are disabled.
```
eugenioenko, line 260: The constructor accepts interceptors instead of an auth provider, and reads can use explicit KAS settings instead of `platformUrl`. The previous requirements were unconditional.

lib/src/opentdf.ts
```diff
@@ -272,7 +272,9 @@ export type TDFReader = {
  *   autoconfigure: false,
  * });
  *
- * const clearText = await client.read({ type: 'stream', location: cipherText });
+ * const clearText = await client.read({
+ *   source: { type: 'stream', location: cipherText },
```
eugenioenko, line 276: `read()` requires `ReadOptions.source`. The previous example passed a `Source` directly and did not type-check.

lib/src/opentdf.ts
```diff
@@ -411,7 +413,7 @@ export class OpenTDF {
     return reader.decrypt();
   }
 
-  /** Closes the OpenTDF instance and releases any resources. */
+  /** No-op retained for API compatibility. */
```
eugenioenko, line 416: The method body is a no-op. The previous comment incorrectly claimed that it released resources.

lib/tdf3/src/client/index.ts
```diff
@@ -336,7 +336,7 @@ export class Client {
 
   /**
    * List of allowed KASes to connect to for rewrap requests.
-   * Defaults to `[this.kasEndpoint]`.
+   * If omitted, decrypt requires `platformUrl` and loads the platform allowlist.
```
eugenioenko, line 346: This property remains undefined unless `allowedKases` is configured. The previous comment incorrectly claimed a default of `kasEndpoint`.

eugenioenko (commented):
Inline notes identify why each previous comment was incorrect.

lib/tdf3/src/client/index.ts
```diff
@@ -378,14 +378,9 @@ export class Client {
...
-   * @param {Object} [config.keypair] - keypair generated for signing. Optional, will be generated by sdk if not passed
-   * @param {String} [config.clientId]
-   * @param {String} [config.kasEndpoint] - Key Access Server url
-   * @param {String} [config.refreshToken] - After logging in to browser OIDC interface user
-   * receives fresh token that needed by SDK for auth needs
-   * @param {String} [config.externalJwt] - JWT from external authority (eg Google)
-   * @param {String} [config.oidcOrigin] - Endpoint of authentication service
+   * Creates a TDF3 client. Requires `kasEndpoint`.
+   * Configure authentication with `interceptors`, `authProvider`, or the legacy OIDC options.
+   * Generates signing keys when `dpopKeys` is omitted.
```
eugenioenko: The previous block documented nonexistent `keypair`, treated required `kasEndpoint` as optional, and omitted current authentication options.

lib/tdf3/src/client/index.ts
```diff
@@ -517,7 +512,7 @@ export class Client {
    * @param [keyMiddleware] - function that handle keys
    * @param [streamMiddleware] - function that handle stream
    * @param [eo] - (deprecated) entity object
-   * @return a {@link https://nodejs.org/api/stream.html#stream_class_stream_readable|Readable} a new stream containing the TDF ciphertext
+   * @returns A decorated stream whose Web ReadableStream contains the TDF ciphertext.
```
eugenioenko, line 529: The return value wraps a Web `ReadableStream`; it is not a Node.js `stream.Readable` as previously documented.

lib/tdf3/src/client/index.ts
```diff
@@ -791,7 +786,7 @@ export class Client {
    * @param params.eo Optional entity object (legacy AuthZ)
    * @param params.assertionVerificationKeys Optional verification keys for assertions.
    * @param params.fulfillableObligationFQNs Optional fulfillable obligation value FQNs (overrides those on the Client)
-   * @return a {@link https://nodejs.org/api/stream.html#stream_class_stream_readable|Readable} stream containing the decrypted plaintext.
+   * @returns A decorated stream whose Web ReadableStream contains the decrypted plaintext.
```
eugenioenko, line 803: The return value wraps a Web `ReadableStream`; it is not a Node.js `stream.Readable` as previously documented.

lib/src/opentdf.ts
eugenioenko: Confirmed as a dead public field: it is declared but never read or forwarded. Assertion signing uses `assertionConfigs[].signingKey`; DPoP uses `dpopKeys`; vendored JOSE helpers do not consume it. Tracked in #1004.

lib/src/opentdf.ts
eugenioenko: This is ignored only by the `OpenTDF` wrapper. Direct `TDF3Client.decrypt()` still honors `concurrencyLimit`; `ZTDFReader.decrypt()` hardcodes 1 instead of forwarding `this.opts.concurrencyLimit`. Tracked as a forwarding bug in #1005.

PR #1031, opened 2026-09-10
fix(sdk): reject GMAC root signatures instead of trusting the manifest (DSPX-4703)

dmihalcik-virtru:
## The bug

A segment's `GMAC` hash is the AES-GCM tag the AEAD just computed over that
segment's ciphertext, so reading the trailing 16 bytes back out is legitimate
— the tag is a real MAC over real data under the real key.

The root signature covers the *aggregate hash*: a concatenation of the segment
hashes. AES-GCM never processed those bytes. Applying the same
trailing-bytes extraction to them returns a keyless copy of the last segment
hash — a value the attacker already has, since it is sitting in the manifest
they are handing you.

`rootSignature.alg` is read from the manifest, which is unauthenticated until
the root signature validates, and unknown values were coerced to `HS256`. So
the attack needs no key at all:

1. Rewrite `rootSignature.alg` to `GMAC`.
2. Truncate, reorder, or duplicate `segments[]` however you like.
3. Recompute the "signature" as the last 16 bytes of the new aggregate hash.
4. `rootIntegrity` reads those bytes back and agrees with itself.

AEAD tags bind no ordering, index, or count — each tag says only "this
ciphertext is intact," never "this is segment 3 of 7." The ordered segment
list is exactly and only what the root signature protects, which is why
losing it costs the whole structural guarantee: a 7-segment file becomes a
2-segment file, or the same segment repeated, and every remaining tag still
verifies.

## The fix

Split the one overloaded `getSignature` into four functions that each say what
they authenticate:

| | HS256 | GMAC |
|---|---|---|
| `segmentIntegrity` | ✅ | ✅ reads the AEAD tag |
| `rootIntegrity` | ✅ | ❌ `IntegrityError` |

plus `segmentIntegrityVersion422` / `rootIntegrityVersion422` for the legacy
hex-then-base64 encoding. After the split there is no longer a code path that
can read a tag out of the aggregate hash, so the misuse isn't a policy check
that could be bypassed — it doesn't exist as a function.

On read, `asRootIntegrityAlgorithm` fails closed: anything other than `HS256`,
in any casing, raises `IntegrityError` instead of being coerced.
`asSegmentIntegrityAlgorithm` stays permissive about casing, since writers in
the wild emit both `GMAC` and `gmac`.

`rootSignature.alg` and `segmentHashAlg` are widened to plain strings in the
manifest types, which is the honest signature: they are attacker-controlled
until validated.

### Don't launder the error we just raised

Raising an `IntegrityError` is only half of it; the caller has to receive one.
`streamToBuffer` drained the plaintext with
`new Response(stream).arrayBuffer()`, and **Chrome replaces *any* error raised
while it pulls a Response body with a bare `TypeError: Failed to fetch`**,
discarding the original. So in the browser — and only in the browser — a
tampered payload arrived at the caller indistinguishable from a dropped
connection. That is exactly the distinction this PR exists to draw: the first
is an attack, the second is a retry.

Draining with a reader instead preserves the error the stream was errored
with. `toString` went through `Response.text()` and had the same problem; it
buffers first now. This is why `DecoratedReadableStream.ts` is in the diff of
an otherwise `tdf.ts`-shaped PR — without it the two payload-tampering tests
below pass under Node and fail under karma.

> **That change is also up on its own as #1035, against `main`.** It is
> orthogonal to the GMAC work — the masking predates it by two years — and it
> is only carried here so this PR's tests can pass before #1035 lands. Once
> #1035 merges, `DecoratedReadableStream.ts` and
> `tests/mocha/unit/decorated-readable-stream.spec.ts` drop out of this diff on
> the next rebase. Review it there, not here.

## How to test

```
cd lib && npm run build && npx mocha 'dist/web/tests/mocha/root-signature.spec.js'
```

`lib/tests/mocha/root-signature.spec.ts` is new — 23 cases in four groups:

- **controls** (6) — a repacked file round-trips; truncation, a hash edit, a
  reorder, and an unforged GMAC downgrade are all caught under HS256.
- **exploit: GMAC root downgrade** (8) — the full keyless attack with
  truncated and with reordered segments, each of the four casings of `GMAC`,
  and an unknown algorithm that must not fall back to HS256.
- **segment integrity is unaffected** (5) — GMAC and HS256 segments both
  round-trip and both still catch payload tampering; lowercase
  `segmentHashAlg` still reads; an unknown segment algorithm is refused.
- **legacy 4.2.2** (4) — hex-then-base64 HS256 still validates, truncation is
  still caught, and the GMAC downgrade is refused there too.

`lib/tests/mocha/unit/decorated-readable-stream.spec.ts` is also new — 7 cases
pinning that `streamToBuffer` concatenates correctly and that both it and
`toBuffer`/`toString` reject with the *identical* error instance the stream
was errored with, never a substitute. It lives under `tests/mocha/unit/` so
webpack picks it up and it runs under karma as well as Node, which matters:
the masking only ever reproduced in Chrome.

### Baseline

Run this PR's tests against #1030's code, i.e. everything here except the
`tdf.ts` fix:

```
14 passing
 9 failing
```

Every case under **controls** stays green, and the 9 failures are exactly the
8 exploit cases plus the lowercase-`segmentHashAlg` case. The exploit failures
are all "expected an IntegrityError" — the forged files decrypt cleanly on
unfixed code. With the fix, all 23 pass.

The `streamToBuffer` fix is pinned the same way: reverting it alone turns 409
karma passes into 6 failures — the two payload-tampering cases above plus four
of the seven error-propagation cases — while Node stays green throughout.

With both fixes the whole suite is green: 409 passing / 6 pending under mocha,
409 under karma, 253 under web-test-runner, coverage thresholds met, and
`tsc --noEmit` and `npm run lint` clean.

## Risk

Crypto read path. Files that were previously accepted and are now rejected are
exactly the ones with a non-HS256 root signature, which had no integrity
guarantee to begin with. Legitimate GMAC-segment files are unaffected — that
path is unchanged and covered above. The write path already refused to produce
a GMAC root as of #1030.

### Interop note

This SDK rejects an absent or empty `rootSignature.alg`; the Go and Java SDKs
default it to `HS256`. Both are safe, since the HMAC still has to verify
either way, and no golden file in xtest exercises it. Worth reconciling, but
not in this PR.

DSPX-4703

Cross-SDK coverage lives in opentdf/tests#594.

## Security

- Strengthened validation of manifest and segment integrity algorithms.
- Root integrity verification now rejects unsupported, invalid, or downgraded algorithms.
- Improved detection of tampering, including altered, reordered, or truncated segments.
- Algorithm validation is case-insensitive where applicable and fails safely for unknown values.

## Compatibility

- Preserved validation support for legacy 4.2.2 files using supported root integrity settings.

## Reliability

- Stream and integrity errors now propagate accurately instead of being masked by generic browser errors.
- Improved buffering and text decoding for streamed content.


lib/tdf3/src/tdf.ts
```diff
@@ -1130,17 +1213,14 @@ async function decryptChunk(
...
+  segmentIntegrityAlgorithm: SegmentIntegrityAlgorithm,
   specVersion: string,
   cryptoService: CryptoService
 ): Promise<DecryptResult> {
-  if (segmentIntegrityAlgorithm !== 'GMAC' && segmentIntegrityAlgorithm !== 'HS256') {
-    throw new UnsupportedError(`Unsupported integrity alg [${segmentIntegrityAlgorithm}]`);
-  }
-  const segmentSig = await getSignature(
-    reconstructedKey, // SymmetricKey (opaque)
+  const segmentSig = await segmentIntegrity(
```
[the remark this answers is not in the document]
dmihalcik-virtru: HS256 is not supported for 4.2.2

lib/tdf3/src/tdf.ts
```diff
@@ -195,14 +195,42 @@ export function isSegmentIntegrityAlgorithm(alg: unknown): alg is SegmentIntegri
...
+  if (!isSegmentIntegrityAlgorithm(alg)) {
+    throw new UnsupportedError(`Unsupported segment hash alg [${alg}]`);
+  }
+  return alg.toUpperCase() as SegmentIntegrityAlgorithm;
+}
+
+/**
+ * Normalize a manifest-declared root algorithm, failing *closed*.
+ *
+ * A ZTDF's `rootSignature.alg` is unauthenticated manifest data. Accepting
```
pflynn-virtru, line 219:
```suggestion
 * A base TDF's `rootSignature.alg` is unauthenticated manifest data. Accepting
```

lib/tests/mocha/root-signature.spec.ts
```diff
@@ -0,0 +1,462 @@
+/**
+ * DSPX-4703 — the root signature is the only thing in a ZTDF that authenticates
```
pflynn-virtru, line 2:
```suggestion
 * The root signature is the only thing in a base TDF that authenticates
```
no jira tickets, some soft guidance that OpenTDF is base TDF

pflynn-virtru (approved):

PR #1042, opened 2026-09-16
fix(sdk): read manifest.json from a TDF archive

pflynn-virtru:
Refs opentdf/platform#3513. The read half of #1040, split out so it can land on its own.

**Not a file-format change.** The writer still emits `0.manifest.json`; this only widens what the reader accepts.

## Problem

The [OpenTDF spec](https://opentdf.io/spec#tdf-structure) names the manifest entry in a `.tdf` archive `manifest.json`:

> The `manifest.json` file MUST be in JSON format and reside within the root of the OpenTDF Zip archive.
>
> — `spec/schema/OpenTDF/manifest.md`

This SDK writes and reads `0.manifest.json`. On the read side it is an exact central-directory lookup with no fallback, so a TDF produced by any implementation written against the published spec is rejected outright — `Unable to retrieve CD manifest`.

The `0.` prefix is a holdover from an early design that anticipated several payload/manifest pairs per archive. That design never shipped. Every SDK agreed on the same wrong name, so nothing caught it.

## Change

`loadTDFStream` and `Client.getPolicyId` resolve the manifest entry through `manifestEntryName()`, which asks for the spec name when the archive carries it and otherwise for the off-spec `0.manifest.json`.

The writer is untouched apart from using the `offspecManifestFileName` constant in place of the string literal, so every archive this SDK produces is byte-identical to before. Renaming what the writer emits is a breaking file-format change that strands existing readers (java-sdk, client-cpp, go, and xtest all look the manifest up by exact name); #1040 carries that half and can land once those consumers are ready. Splitting the read side off lets this SDK start accepting spec-conforming archives now, with no downstream coordination.

The dual-name handling lives in `tdf.ts`, not in `ZipReader`. `ZipReader`'s signature is unchanged — its other lookup is a payload path that must stay exact, and keeping the name a parameter leaves the generic reader generic.

Resolution runs against the **central directory**, not against a failed read. That keeps other failures honest: an oversized manifest under the spec name stays a size error rather than silently yielding a superseded off-spec entry. The tests pin this specifically — see below.

The name the writer emits is a separate binding, `manifestEntryNameToWrite`, from the two names the reader accepts. Repointing it is the whole of the follow-up writer change, and it cannot disturb the read side — collapsing the two roles into one constant is how a spec-name flip would silently strand every archive earlier releases wrote.

Preferring the spec name when an archive carries **both** entries has a tampering cost, taken knowingly; see [the review-notes comment](#issuecomment-5703189131) for that and for the `0.payload` gap.

`0.payload` is untouched — it is recorded in the manifest's `payload.url`, so renaming it would alter manifest contents rather than just archive layout.

## Testing

New `lib/tests/mocha/manifest-entry-name.spec.ts`, 16 tests:

| Group | What it pins |
|---|---|
| `manifestEntryName` | spec name when present; off-spec name when the spec name is absent; spec name preferred when an archive carries both; off-spec name when it carries neither |
| `writer` | the central directory of a freshly encrypted archive is still exactly `['0.payload', '0.manifest.json']`; that archive round-trips |
| `spec-named archives` / `off-spec-named archives` | generated from one table, so each name is covered explicitly rather than by whichever one the writer happens to emit: the fixture really carries that name, it decrypts, and its policy id equals the original archive's |
| `an oversized spec-named entry shadowing a valid off-spec one` | `decrypt` and `getPolicyId` both fail on size rather than serving the valid off-spec manifest |
| `archives with no manifest entry under either name` | `decrypt` reports `Unable to retrieve CD manifest` |

Fixtures are archives this SDK wrote, read back entry by entry and rebuilt with the manifest entry renamed. The two names differ in length, so the local headers, central-directory offsets and the EOCDR are all rebuilt rather than patched, and the payload bytes and manifest content are carried over unchanged. `renameManifestEntry` throws unless exactly one entry looked like a manifest, so a fixture can never quietly come back unrenamed. What the fixtures pin is name resolution and nothing else: they are this SDK's zip dialect throughout (zip64, data descriptors, STORE), so a producer with a different dialect exercises `parseCDBuffer` paths these tests do not reach.

Mutation-tested three ways:

| Mutation | Result |
|---|---|
| `manifestEntryName` returns `offspecManifestFileName` (behavior before this PR) | **6 failing** — both spec-name unit tests, the spec-named archive behaviors, and both shadowing tests |
| reader becomes `try { getManifest(spec) } catch { getManifest(off-spec) }` | **2 failing** — both shadowing tests. This is the design the implementation comment rejects, and an earlier revision of these tests missed it entirely |
| `manifestEntryNameToWrite` repointed at the spec name (i.e. #1040 lands on top) | **1 failing** — only the deliberate writer pin. All read coverage, including the off-spec group, stays meaningful |
| `offspecManifestFileName` collapsed onto the spec name (the mistake the split above exists to prevent) | **5 failing** — the writer tripwire, both off-spec unit tests, and the whole off-spec read group. An earlier revision of these tests, written in terms of the constants, missed this one |

Assertions spell the two entry names out as literals rather than importing the constants: a suite written in terms of the constants it is testing cannot notice them changing, which is exactly how the last mutation above used to pass.

```
$ cd lib && npm run test:mocha   # exit 0 — 395 passing
$ cd lib && npm run test:wtr     # exit 0 — 253 passed, 15 files
$ cd lib && npm run lint         # exit 0
$ cd lib && npx prettier --check "{src,tdf3,tests}/**/*.ts"   # clean
```

**Not run:** `cli`/`web-app` suites (untouched by this change), `web-app/tests` (Playwright) and xtest, which need a live platform.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved TDF archive compatibility by recognizing the standard `manifest.json` entry name while continuing to support archives using `0.manifest.json`.
  - Policy identification and decryption now work consistently across supported manifest naming conventions.
  - Archives with oversized or missing manifest entries now return appropriate retrieval or size-related errors.
  - TDF archives created by the SDK continue to use the existing manifest format.


pflynn-virtru:
## Review notes: two tradeoffs taken deliberately

Recording these so they read as decisions rather than oversights.

### Preferring `manifest.json` widens one tampering surface

`manifestEntryName` prefers the spec name when an archive carries both entries. Combined with `ZipReader.getCentralDirectory`, which signature-scans only the last 1000 bytes and never cross-checks the EOCDR's record count or CD offset, that means an attacker can **append** to a genuine archive this SDK wrote — leaving every original byte intact as a prefix — a forged manifest, one extra central-directory record named `manifest.json`, and a fresh EOCDR. `manifestEntryName` prefers the appended name, and `getManifest`'s `.find()` returns it.

On `main` the hardcoded `0.manifest.json` lookup hit the original record first, so this is a real if narrow regression. Scope:

- **Not** plaintext disclosure. Decryption is still gated by the KAS policy binding, so a forged policy fails at rewrap.
- It does reach every pre-decryption manifest consumer: `Client.getPolicyId`, `OpenTDF` reader `.manifest()`, CLI `inspect`, the web-app KAO inspector, and the `keyAccess[].url` used to pick the rewrap target (guarded by `OriginAllowList` unless `ignoreAllowlist` is set).

Taken knowingly. Preferring the name this SDK writes would close it, at the cost of flipping the preference again when the writer moves to the spec name. The underlying weakness is the 1000-byte central-directory window, which is pre-existing and worth replacing with a real EOCDR parse (record count + CD offset + size) independently of this PR — the same window also makes `adjustHeaders` follow `relativeOffsetOfLocalHeader` without validating the local header signature or that its stored name matches the CD record.

### Spec-named archives still need `0.payload`

`getPayloadSegment` is called with the literal `0.payload`, and `manifest.payload.url` is read nowhere in this codebase. Real spec-conforming archives are unaffected — platform#4049 deliberately leaves `0.payload` alone — but an archive that both uses the spec manifest name *and* names its payload something else now gets past `loadTDFStream`, completes a KAS rewrap, and fails only when the stream is drained. Before this PR it failed locally with no network. Honouring `payload.url` is the separate fix; noting it here as a known gap.

### Also raised, deliberately left alone

`getManifest` does not check `compressionMethod`, so a DEFLATE'd manifest from a foreign producer escapes as a bare `SyntaxError` instead of `InvalidFileError`. Newly reachable, pre-existing, one-line fix — but it belongs with the zip-reader hardening above rather than in a naming change.

pflynn-virtru:
## Pre-review findings

Three behaviors this PR introduces that the suite doesn't cover. Recording rather than fixing here — #1044 carries the fix for the first, stacked on this branch.

**1. A spec-named entry is selected on presence alone, with no shape check.** Adding a third entry named `manifest.json` to an SDK-written archive:

| Content | `decrypt` / `getPolicyId` |
|---|---|
| `{"name":"my-web-app"}` | `TypeError: Cannot read properties of undefined` |
| `hello` | `SyntaxError: Unexpected token 'h'` |
| zero-length | `SyntaxError: Unexpected end of JSON input` |

`instanceof InvalidFileError === false` in all six cases, so callers catching `TdfError` fall through. All three decrypt on `main`, where the hardcoded lookup ignored the extra entry.

**2. `getPolicyId` gained a `this` dependency.** The old body touched only module-scope helpers. Routing through `this.loadTDFStream(...)` means `const { getPolicyId } = client` and `sources.map(client.getPolicyId)` now throw. Public surface; nothing in `cli/` or `web-app/` calls it, so CI can't see it.

**3. `it('reads the policy id')` is vacuous.** It asserts two `getPolicyId` calls agree with each other, so `assert.equal(undefined, undefined)` satisfies it — verified by patching the built method to return a nonexistent field: 16/16 still green. It's the only assertion on a policy-id value in the repo.

**On the both-names preference.** Worth weighing: *reject* an archive carrying both names rather than preferring either. No legitimate producer emits both, it closes the appended-CD-record surface without a preference that has to flip when the writer moves, and it costs one condition.

Also unchanged for foreign archives: `0.payload` is hardcoded (`manifest.payload.url` has zero read sites in `lib/`), the central directory is parsed from a 1000-byte tail window, and DEFLATE is unsupported. Pre-existing, but the reason "reads spec archives" is narrower than it sounds.

**Testing.** Merged `main` (picking up #1031, which rewrote 181 lines of `tdf3/src/tdf.ts`) and re-ran: 425 passing, 0 failing. Hunks don't overlap.

dmihalcik-virtru (approved):

